PR #1, opened 2015-06-07
Implement xhr error handling

m0ppers:
The backend might return errors which should be handled on the frontend.
In my case the backend immediately tries to check the file type of the
upload and will throw a 417 HTTP with some error info in the body. There
is no possibility to handle that error right now.

### DIFF
--- a/dist/tus.js
+++ b/dist/tus.js
@@ -113,9 +113,7 @@ var Upload = (function () {
     }, {
         key: "_emitXhrError",
         value: function _emitXhrError(xhr, err) {
-            err.status = xhr.status;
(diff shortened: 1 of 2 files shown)


Acconut:
Thanks for your contribution. While I think this is very useful, we should expose the entire XMLHttpRequest object to allow inspecting the request and response headers and so on.
What about:
```js
    _emitXhrError(xhr, err) {
        err.originalRequest = xhr
        this._emitError(err)
    }
```

m0ppers:
Makes total sense to me :)

Acconut:
Thanks, I will add a test and then merge this. :+1:

PR #5, opened 2015-09-29
Dist: add minified version

fahad19:
## New commands

* `npm run minify` to generate `tus.min.js` file
* `npm run dist` to browserify and minify together

(PR also includes updated dist files as a separate commit, which can be ignored)


Acconut:
Thanks :)

PR #11, opened 2015-10-08
ESLint

fahad19:
## Things done 

* Added ESLint config (Run `npm run lint`)
* Updated `/lib` directory for passing linters
* Dist files updated
* `.editorconfig` file added

## Before merge

I tried running `npm test` but it didn't work for me, so can't verify if I broke something or not.


lib/upload.js
```diff
@@ -1,262 +1,280 @@
...
-    resume: true,
-    onProgress: null,
-    onSuccess: null,
-    onError: null,
-    headers: {},
-    chunkSize: Infinity,
-    withCredentials: false
-}
+/* global window */
+import fingerprint from './fingerprint';
```
Acconut: Is there a reason why single and double quotes are mixed?

fahad19: nope, no reason. we should stick to only one way. made it single quotes in https://github.com/fahad19/tus-js-client/commit/e5f80c0889a386751624040f62607372f01da22d

PR #14, opened 2015-10-21
Improvements to demo page

abkaplan07:
We liked using the demo page as a quick way to test and validate our TUS server. This pull request has the following changes to make it more flexible:
- Endpoint can be set on the page.
- Chunk size can be configured, allowing files to be uploaded by multiple PATCH requests.


Acconut:
Thanks, this is convenient. I had to cast the string to a number, thou, to make it work properly. :)

PR #35, opened 2016-05-21
Add abality to pass params into the request

canaandavis:
Any interest in something like this?

I've found that working with some 3rd party apis, such as the Youtube Api, requires that certain params be passed through the request body.  So we are unable to take advantage of the metadata header.

This way we would have the option to send params when we first create the upload.


Acconut:
First of all: Thank you for this contribution.

> I've found that working with some 3rd party apis, such as the Youtube Api, requires that certain params be passed through the request body. So we are unable to take advantage of the metadata header.

Frankly, I cannot follow your argument on why using the metadata headers is not sufficient for your use case. Is the information you want to attach so big that you are unable to transport it using headers? Could you please elaborate on this a bit more?

In addition, I am not entirely in favor of your implementation. It limits the body to a JSON-encoded structure but – if this feature is added – tus.Upload should accept a generate object which is then directly passed to `send()` to allow using `Blobs` and other types, too, just as `XMLHttpRequest#send()` does.

canaandavis:
Thanks for getting back with me!

This was probably the wrong approach to present this.  I should have opened an issue as this was meant to be more of a question about the ability / interest in being able to pass information directly to `send()`.  The implementation I presented was very limited in scope, so I apologize about that.

The issue that I was running into with the Youtube Api is that I am interfacing directly with their server and I need to pass some information about a video.  That information may look something like this.

`{
  snippet: {
    title: 'video title',
    description: 'video description'
  },
  status: {
    privacyStatus: 'unlisted'
  }
}`

The api is looking for this information to be explicitly in the request body, I also don't think the current implementation of `encodeMetadata` is handling nested objects like this. 

Let me know if that information is helpful, and again sorry for submitting this here, I realize now this was not the best spot for that.

Thanks for the awesome library, and let me know your thoughts on that, maybe I'm just missing something here.

Acconut:
Thank you a lot for these kind words and there is no need for you to apologize for anything. :)
Before we talk more about this specific feature request, I believe that you try to use tus-js-client with a general misconception:

> The issue that I was running into with the Youtube Api is that I am interfacing directly with their server and I need to pass some information about a video. 

If you say that you are *directly interfacing with the YouTube API*, I understand that you are using no special tus server between your app and the remote API. And that's not how it actually works. tus-js-client can only communicate with servers which do understand the tus protocol. An example is the [tusd](https://github.com/tus/tusd) server but YouTube does not accept uploads using the tus protocol (for now, maybe they will hop on the train sometime) and therefore *directly interfacing with the YouTube API* is not possible.

Please correct me, if I had a wrong thought here.

canaandavis:
> tus-js-client can only communicate with servers which do understand the tus protocol.

Thank you!  This is what I was missing, makes much more sense now.  Really appreciate your time, I will take a look at the tusd server.

I don't know how applicable this request is any longer, I will need to familiarize myself with the tus protocol.  Again thank you for the clarity!

Acconut:
That's great to hear. I will close this discussion for now but feel free to reopen if you have more insights.

PR #48, opened 2016-09-22
Better stream handling (WIP)

olalonde:
I made the StreamSource code smaller and easier to understand. I also
removed the need to pass a "chunkSize" argument. It should also now
support readable streams that are not necessarily instances of Readable
which is quite common. See https://github.com/sindresorhus/is-stream


Acconut:
I definitely like the direction where this is headed. Thank you!

olalonde:
@Acconut thanks! Do you know why "chunkSize" was a required argument for stream sources? Removing it hasn't been a problem for me when testing with streams but I wonder what was the original reason it was there.

Acconut:
The previous stream implementation always kept the chunk which is currently uploaded in memory. This buffer will then be used if the chunk needs to be uploaded again because there was an error in the transmission. In order to properly size this buffer, it needed the chunkSize. I am not sure if your updated version includes this behavior which is actually quite important.

olalonde:
I see, and no it doesn't :( I assume the default value for chunkSize is `Infinity` so that it tries to upload the whole thing in one shot? That could be a lot of data to keep in memory for large streams (e.g. multiple GBs). May I suggest instead that we instead pipe the stream in parallel to a temporary file? And when retrying if the offset is in that file, start from there and continue with the original stream. Does that make sense?

Acconut:
> I assume that the default value for chunkSize is Infinity so that it tries to upload the whole thing in one shot?

As you can see from the original code (https://github.com/tus/tus-js-client/pull/48/commits/8347ecffcb29c6014b818e75fc59ffc99c37675f#diff-fbf9574e6ad4d6644955156e0062e4b1L44), it rejected `Infinity` as a value for chunkSize if you try to upload a stream.

> May I suggest instead that we instead pipe the input stream in parallel to a temporary file? And when retrying, if the offset we need is in that file, start from there and continue with the original stream once we read the entire file from the offset. Once the upload has succeeded the temporary file could be removed. That would also remove the requirement of having to specify an arbitrary chunk size.

My first implementation used this approach. While it could simplify the code drastically, it introduced additional overhead which gets especially noticeable if you upload huge files. There the entire upload would be written to disk, while the current approach just keeps the current chunk in memory.

> Another alternative would be to allow a "stream returning function" as input and just get a new stream if the upload fails unexpectedly.

I doubt that this plays well with the pipe chains which are quite common in Node land. Furthermore, I imagine situations where this will involve a lot of calculations which need to be redone. Also there may be cases in which it's simply impossible to generate the exact same stream.

Acconut:
Closing this PR as work has been discontinued at that stage. We can reopen it, of course.

Issue #52, opened 2016-11-20
Transpile Job on macOS: sed is not portable


mmathys:
The transpile job is currently set like this:

```bash
babel -d lib.es5/ --presets es2015 lib/ && sed -i '1i // Generated by Babel' lib.es5/*.js lib.es5/*/*.js
```
However macOS machines have a different `sed` command: they're not using the GNU version, they're using another version.

[It's not possible to use the same sed command on both Linux and macOS](http://stackoverflow.com/a/4247319/2022821).

The version above only runs on Linux, on macOS it throws `sed: 1: "lib.es5/error.js": extra characters at the end of l command`.

As of now we're relying on contributors with Linux to transpile the code to es5.

mmathys:
🎉

PR #53, opened 2016-11-20
Fix tus.isSupported on iOS in Private Mode

### LINKED ISSUE #49: iOS private mode: Missing check for tus.isSupported.
I recently encountered an issue where the value of `tus.isSupported` was set to `true` but the browser didn't support tus:

The private mode of Safari on iOS doesn't allow access to the local storage ([see this thread](http://stackoverflow.com/questions/21159301/quotaexceedederror-dom-exception-22-an-attempt-was-made-to-add-something-to-st)).

Therefore it might be good to implement a check which checks whether the local storage is accessible, [something like this maybe](http://stackoverflow.com/questions/16427636/check-if-localstorage-is-available).

mmathys:
The transpile job wasn't executed yet, see #52


lib/browser/storage.js
```diff
@@ -5,10 +5,18 @@ try {
   hasStorage = "localStorage" in window;
   // Attempt to access localStorage
   localStorage.length;
```
Acconut: I believe this access could be removed now, as it's superseded by our additions.

lib/browser/storage.js
```diff
@@ -5,10 +5,18 @@ try {
...
+
+  // Attempt to store and read entries from the local storage to detect Private
+  // Mode on Safari on iOS (see #49)
+  var test = 'tusSupport';
+  localStorage.setItem(test, test);
+  localStorage.removeItem(test);
```
mmathys: When test doesn't exist in the first place you'd attempt to set and item with the value `undefined`, then an error would be thrown even when there was local storage support

mmathys: Also, I have the feeling that the probability that an attribute with `tusSupport` is used by another application is very small.

Alternatively, we could use something like `'tusSupport_dHVzU3VwcG9ydA=='`

mmathys:
Okay @Acconut. If you say so, I'll change it in a bit

mmathys:
Didn't transpile anything yet. Didn't know how to merge your @Acconut commit back into my fork.

Acconut:
Great, thank you! Would you also mind addressing the linting issues (see https://travis-ci.org/tus/tus-js-client/jobs/178254879#L406)?

Acconut:
> Didn't know how to merge your @Acconut commit back into my fork.

You should be able to pull from master into your local fork using `git pull https://github.com/tus/tus-js-client.git master`. This will include the necessary commits which make the transpile step work on macOS.

Acconut:
Has been merged in https://github.com/tus/tus-js-client/commit/19594295025114603bb7757491252316c8e167f3. Thank you for helping with this!

PR #69, opened 2017-05-09
Don't remove fingerprint on server errors (fixes #68)

### LINKED ISSUE #68: Don't remove fingerprint on server errors?
Hi,

in [upload.js#L314](https://github.com/tus/tus-js-client/blob/master/lib/upload.js#L314) we trigger a `Storage.removeItem(this._fingerprint);` even on server errors (`statusCode >= 500`).
But probably there's only a short service outage and it could be resumed after the backend is up again. 

I think we should only remove the stored fingerprint on client errors (< 500). What do you think?

psi-4ward:


Acconut:
I apologize for taking so long with this, I lost it from my radar. Thank you very much!

PR #70, opened 2017-05-09
Don't retry on BadRequests/Client-Errors (fixes #66)

### LINKED ISSUE #66: abort retry on BadRequests
Hi,

it's does not make sense to retry if the server responds with `400 <= statusCode < 500`.
In this case there is something wrong with the request on the client side and we should fire the error callback.

psi-4ward:


Acconut:
I apologize for taking so long with this, I lost it from my radar. Thank you very much!

Issue #62, opened 2017-03-01
Support Resumability for nodejs environment


ifedapoolarewaju:
A number of use cases where this feature is desired seem to have come up ([uppy](https://github.com/transloadit/uppy) and [uppy-server](https://github.com/transloadit/uppy-server) being an example).  So maybe this is a great time to have this feature implemented? cc @Acconut @kvz

Acconut:
There's what I propose:

We add a `urlStorage` property to the options which must be null or an instance with following layout:

```js
class UrlStorage {
  setItem(fingerprint, url, callback)
  getItem(fingerprint, callback)
  deleteItem(fingerprint, callback)
}
```

Notice how the API designed with callbacks to allow an asynchronous usage (which is required for Redis etc). However, we should also consider using Promises as this is probably going to be the standard in new browser APIs.

Furthermore, if the `urlStorage` options is null (it's default value) and we are in the Node environment, no resumability should be enabled. However, if we are in a browser, it should be set to an UrlStorage instance which wraps the localStorage API. This ensure backwards-compatibility.

What do you think?

kvz:
I like this proposal, I do think it's nice that on the server, it would be cool if you could enable reusability with just configuration, so that not everybody has to write the same adapter for local file storage. I'm thinking something like this on the server:

```js
const fsAdapter = require('tus-js-client/storage-adapters/fs')
const opts = {
   urlStorage: {
     adapter: fsAdapter,
     conf: {
        dir: '/tmp/tus-resumability'
     }
   } 
}
``` 

In the browser, if we detect the default `null` as `urlStorage`, we substitute with this internally under the hood, much like you said, Marius:

```js
if (opts.urlStorage === null && env === 'web') {
  const webAdapter = require('./storage-adapters/browserLocalStorage')
  opts.urlStorage = {
     adapter: webAdapter,
  }
}
```

Acconut:
> so that not everybody has to write the same adapter for local file storage

Sure, that's what I also thought. We should provide a basic file-based url storage.

Regarding your code snippet I would prefer if the option interface looked more like:

```js
// my-server-app.js
const FileUrlStorage = require('tus-js-client/storages/file')
const opts = {
   urlStorage: new FileUrlStorage({
      dir: '/tmp/tus-resumability'
   })
}

For me it's easier to see what's going on if you pass an object instead of a constructor and the corresponding arguments for the constructor.

kvz:
💯% agreed!

[...]

PR #73, opened 2017-05-23
feature: Support Resumability for node env #62

ifedapoolarewaju:
I'd like to get some feedback on this before proceeding to tests.

P.S we might need to rewrite the default `fingerprint` module. Or maybe dedicate one to the `node` package and another to the `browser` package. As of now the `fingerprint` can't get a file `size`, `name`, `type` and `lastModified` on the node env. So it returns just `tus---` as a fingerprint.


lib/browser/storage.js
```diff
@@ -22,17 +22,25 @@ try {
 
 export const canStoreURLs = hasStorage;
 
-export function setItem(key, value) {
-  if (!hasStorage) return;
-  return localStorage.setItem(key, value);
-}
+export class DefaultStorage {
+  setItem (key, value) {
+    return new Promise((resolve, reject) => {
```
Acconut: We are currently using callbacks for the event system. What do you think if we do not use Promises to keep this consistency?

lib/browser/storage.js
```diff
@@ -22,17 +22,25 @@ try {
 
 export const canStoreURLs = hasStorage;
 
-export function setItem(key, value) {
-  if (!hasStorage) return;
-  return localStorage.setItem(key, value);
-}
+export class DefaultStorage {
```
Acconut: A more detailed name, such as `LocalStorage` or `WebStorage` (?), would be appropriate.

lib/node/storage.js
```diff
@@ -1,15 +1,57 @@
 /* eslint no-unused-vars: 0 */
+import {readFile, writeFile} from "fs";
 
-export const canStoreURLs = false;
+export const canStoreURLs = true;
 
-export function setItem(key, value) {
+export class DefaultStorage {
```
Acconut: A more detailed name, such as `FileStorage`, would be appropriate.

lib/node/storage.js
```diff
@@ -1,15 +1,57 @@
...
+  constructor (dir) {
+    this.dir = dir
+  }
 
-}
+  setItem (key, value) {
+    return new Promise((resolve, reject) => {
+      this._getData().then((data) => {
```
Acconut: Unfortunately, this is not secured against race conditions as this is not an atomic operation. Think of a situation where between the _getData and _writeData calls another node process modifies the file. Those changes would be lost. Maybe there is a module on NPM which already provides a simple file-based key-value storage?

lib/browser/storage.js
ifedapoolarewaju: Sure! I think I misunderstood your comment here:

> Promises as this is probably going to be the standard in new browser APIs

That's why I went with the Promise approach. Will make changes 👍

ifedapoolarewaju:
@Acconut do you mind reviewing this again whenever you get the chance? I have addressed the issues you raised from your last review.

PS. this change is required for Uppy's support for ReactNative's resumability.
cc @kvz @arturi

lib/upload.js
```diff
@@ -191,27 +199,33 @@ class Upload {
...
-
-    // An upload has not started for the file yet, so we start a new one
-    this._createUpload();
   }
 
   abort() {
     if (this._xhr !== null) {
       this._xhr.abort();
       this._source.close();
-      this._aborted = true;
```
ifedapoolarewaju: I figure this was a bug, since it should always set `_aborted` to true after `abort` is called

lib/fingerprint.js
```diff
@@ -14,3 +20,28 @@ export default function fingerprint(file, options) {
     options.endpoint
   ].join("-");
 }
+
+function reactNativeFingerprint(file, options) {
+  let exifHash = file.exif ? hashCode(JSON.stringify(file.exif)) : "noexif";
```
Acconut: What does `hashCode` exactly do? What kind of hash does it calculate? Also what does `file.exif` look like?

lib/node/storage.js
```diff
@@ -1,15 +1,98 @@
...
+import * as path from "path";
 
-export const canStoreURLs = false;
+export const canStoreURLs = true;
+const defaultStoragePath = path.join(process.cwd(), "tus-url-storage.json");
 
-export function setItem(key, value) {
+export class FileStorage {
+  constructor(filePath) {
+    this.path = filePath || defaultStoragePath;
```
Acconut: I don't think there should be a default storage path. Instead a user should always explicitly name the path that should be used.

lib/node/storage.js
```diff
@@ -1,15 +1,98 @@
...
+      }
+
+      release().then(cb).catch(cb);
+    };
+  }
+
+  _writeData(data, cb) {
+    const opts = {
+      encoding: "utf8",
+      mode: 0o666,
```
Acconut: `0666` as file permission seems a bit too open for me. 0664 should be enough or do you disagree?

lib/node/storage.js
```diff
@@ -1,15 +1,98 @@
...
+        }
+
+        delete data[key];
+        this._writeData(data, (err) => cb(err));
+      });
+    }).catch(cb);
+  }
 
-export function removeItem(key) {
+  get _lockfileOptions() {
```
Acconut: Please don't use private getters for static objects. I think that's causing more confusion than it's bringing benefit.

lib/node/storage.js
```diff
@@ -1,15 +1,98 @@
...
+  }
+
+  _getData(cb) {
+    readFile(this.path, "utf8", (err, data) => {
+      if (err) {
+        // return empty data if file does not exist
+        err.code === "ENOENT" ? cb(null, {}) : cb(err);
+        return;
+      } else {
+        data = !data.trim().length ? {} : JSON.parse(data);
```
Acconut: `JSON.parse` can throw an error if the parsing fails. That error should be caught to avoid crashing the app.

lib/node/storage.js
```diff
@@ -1,15 +1,98 @@
...
+        cb(null, data);
+      }
+    });
+  }
 }
+
+export function getStorage() {
```
Acconut: In a Node.js environment there must not be a default URL storage. Otherwise it would be a breaking API change which is not allowed in non-major releases. I think it's best if the user explicitly opts into using the FileStorage.

lib/upload.js
```diff
@@ -191,27 +199,33 @@ class Upload {
     }
 
     // Try to find the endpoint for the file in the storage
-    if (this.options.resume) {
+    if (this.options.resume && this._storage) {
```
Acconut: `this.options.resume && this._storage` is repeated quite often. Please use a private property such as `this._hasURLStorage` instead.

lib/fingerprint.js
ifedapoolarewaju: the `file.exif` is an object containing various properties about the file. So we are hashing the entire content for uniqueness.

ifedapoolarewaju:
> also have a look at the Travis tests which are failing for older Node.js versions.

the test is broken for older versions because `node-properlockfile` does not support node LTS versions that have reached EoL. So I wonder if tus-js-client needs to support these older versions too? Should we rather make this a major version release, and not support unmaintained older versions of node?

lib/node/storage.js
kvz: How about not letting anyone read the files either? So `660`?

Acconut:
> the test is broken for older versions because node-properlockfile does not support node LTS versions that have reached EoL. So I wonder if tus-js-client needs to support these older versions too? Should we rather make this a major version release, and not support unmaintained older versions of node?

We cannot drop support for older versions without a major release. There are plans for a major release but I am not too keen on doing a major release just for this feature. Is there no way to use another library which is compatible?

lib/fingerprint.js
Acconut: Yes, but what kind of hash? It's obviously not a cryptographic one. Why did you choose to imitate Java's `hashCode`. It would be nice if you shared a bit of your thinking behind those decisions.

ifedapoolarewaju:
> We cannot drop support for older versions without a major release. There are plans for a major release but I am not too keen on doing a major release just for this feature. Is there no way to use another library which is compatible?

There aren't many lockfile libraries for nodejs. An alternative is to downgrade the version of the lockfile package from 4.x to 2.x which supports older node versions.
But I would rather vote for a major release tbh. Relying on a far older version of a library deprives us of improvements that might have been made on more recent versions.

lib/fingerprint.js
ifedapoolarewaju: tbh there was little consideration from my end. The snippet was recommended by @kvz during a pairing session, and I simply rocked with it.

ifedapoolarewaju:
@Acconut all issues addressed

lib/fingerprint.js
ifedapoolarewaju: @Acconut is there a different approach you would rather recommend here?

Acconut:
@ifedapoolarewaju Please stop force pushing, it is making incremental reviews impossible and I basically have to review everything from the beginning again. Besides that there is no need for a force push since we squash every PR anyways.

lib/upload.js
Acconut: Good catch! That would not have been a problem before this PR but might causes issues now :+1:

lib/upload.js
```diff
@@ -363,10 +387,14 @@ class Upload {
 
     xhr.onload = () => {
       if (!inStatusCategory(xhr.status, 200)) {
-        if (this.options.resume && inStatusCategory(xhr.status, 400)) {
+        if (this.options.resume && this._storage && inStatusCategory(xhr.status, 400)) {
```
Acconut: You can use `_hasStorage` here.

test/spec/upload.js
```diff
@@ -54,8 +66,9 @@ describe("tus", function () {
 
     it("should upload a file", function (done) {
       var file = new Blob("hello world".split(""));
+      var host = "http://files.tus.io";
```
Acconut: Would you mind explaining the reasoning behind this change? I am usually a fan of DRYing up code but in tests I prefer to repeat constants. This makes adapting tests a bit harder but ensures that every change happens on purpose.

test/spec/upload.js
```diff
@@ -169,50 +184,55 @@ describe("tus", function () {
...
-      req.respondWith({
-        status: 201,
-        responseHeaders: {
-          "Location": "//localhost/uploads/foo"
-        }
-      });
+        req.respondWith({
+          status: 201,
+          responseHeaders: {
+            "Location": "//relative.tus.io/uploads/foo"
```
Acconut: By changing the Location header you removed on of the purposes of this test. Please change it back, so that the hostname from the Location header is different than the one from the endpoint.

test/spec/upload.js
```diff
@@ -169,50 +184,55 @@ describe("tus", function () {
       done();
     });
 
-    it("should resolve relative URLs", function () {
+    it("should resolve relative URLs", function (done) {
       var file = new Blob("hello world".split(""));
+      var host = "http://relative.tus.io";
       var options = {
-        endpoint: "http://master.tus.io:1080/files/"
```
Acconut: Please add the port back, it's important for this test case.

test/spec/upload.js
```diff
@@ -14,6 +14,18 @@ if (isNode) {
   var Blob = Buffer;
 }
 
+var getStorage = function () {
+  // this storage is only needed for the node environment. It defaults to localStorage
+  // for browser.
+  if (isNode) {
+    var temp = require("temp");
+    var storagePath = temp.path();
+    return new tus.FileStorage(storagePath);
```
Acconut: Please add code to empty the storage after every test to ensure that each unit test is run in isolating. This is already done for localstorage in the browser: https://github.com/tus/tus-js-client/pull/73/files#diff-893840c5db7894faca0e2e7ea5a7020dR47

lib/browser/storage.js
```diff
@@ -22,17 +22,23 @@ try {
...
+    if (!hasStorage) return cb();
+    cb(null, localStorage.removeItem(key));
+  }
 }
 
-export function removeItem(key) {
-  if (!hasStorage) return;
-  return localStorage.removeItem(key);
+export function getStorage() {
+  return new LocalStorage();
```
Acconut: Instead of checking for `!hasStorage` in every method of LocalStorage we can now check for it inside `getSource`, i.e.:

```
getStorage() {
  if(!hasStorage) return null;
  return new LocalStorage();
}
```

lib/node/storage.js
```diff
@@ -1,15 +1,111 @@
...
+        retries: 5,
+        minTimeout: 20
+      }
+    };
+  }
+
+  _releaseAndCb(release, cb) {
+    return (err) => {
+      if (err) {
+        // @TODO consider combining error from release callback
```
Acconut: Can you address this TODO?

lib/fingerprint.js
Acconut: I don't have a concrete alternative but I prefer to know the intent behind code like this. So if we want to change it in the future we know what to look out for.

test/spec/upload.js
ifedapoolarewaju: there's no need to do this for `FileStorage` because we are always calling `var temp = require("temp");` which creates a separate storage file.

test/spec/upload.js
ifedapoolarewaju: the primary reason for this change was not necessarily to keep the code DRY, but it was so to make the host of each test case unique. The reason is so that the [nextTick function](https://github.com/tus/tus-js-client/pull/73/files/e2ceee428e0a35fa8bad046a71883888c88ab6b3#diff-893840c5db7894faca0e2e7ea5a7020dR1062) can correctly identify which new request belongs to a given test case.

And since the values of the `host` was changing by each test case, it became hard to repeat the correct host value within the test case

lib/upload.js
Acconut: Addressed in https://github.com/tus/tus-js-client/pull/152.

lib/node/storage.js
Acconut: Addressed in https://github.com/tus/tus-js-client/pull/152.

test/spec/upload.js
Acconut: Ok, fair enough.

test/spec/upload.js
Acconut: Addressed in https://github.com/tus/tus-js-client/pull/152.

test/spec/upload.js
Acconut: Addressed in https://github.com/tus/tus-js-client/pull/152.

lib/browser/storage.js
Acconut: Addressed in https://github.com/tus/tus-js-client/pull/152.

test/spec/upload.js
Acconut: Ok, so we have to remember to use a unique URL for every test case now if we use waitTillNextReq? If so, could you add this as a comment to the tests?

PR #74, opened 2017-05-30
remove obsolete `babel` dependency

goto-bus-stop:
`babel` no longer does anything since Babel v6.

npm was installing the `babel` executable stub from the `babel` package instead of the compiler executable from the `babel-cli` package for me, so `npm run transpile` would fail. This removes the obsolete dependency so the `babel-cli` version will always be installed instead.


Acconut:
Thank you :)

PR #75, opened 2017-05-30
fix buffer initialization in base64 encoding in node

goto-bus-stop:
If a number was passed to `encode()`, the buffer would be created with
uninitialised memory. This patch casts anything that's passed in to
a string first and then uses the safe `Buffer.from` API. `Buffer.from`
was added in Node v4 but the `buffer-from` module ponyfills it for older
Node versions.

See [nodejs/node#4660](https://github.com/nodejs/node/issues/4660)

### DIFF
--- a/lib/node/base64.js
+++ b/lib/node/base64.js
@@ -1,4 +1,4 @@
-import bufferFrom from 'buffer-from';
+import bufferFrom from "buffer-from";
 
 export function encode(data) {
   return bufferFrom(String(data)).toString("base64");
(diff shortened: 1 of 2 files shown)


Acconut:
Ouch, it hurts seeing such errors arise in JavaScript. Thank you a lot for solving this. Could you please address the one linting issue? https://travis-ci.org/tus/tus-js-client/jobs/237541799#L428

goto-bus-stop:
Done :)

Acconut:
Thank you, I added a test and merged in 66694b4a93c8f9a7cda5ee4ea52e3692df926ef4

PR #76, opened 2017-06-07
Add onLoadStart constructor option

gavboulton:
Add an onLoadStart constructor option that gets called when an upload starts. This uses the XMLHttpRequest.onloadstart so will only work in browsers.


Acconut:
First of all, thank you for this PR, this is always greatly appreciated. However, I would also like to know your reason for such a feature. Would you mind explaining your background a bit? Maybe there is already a way to achieve what you wanted.

Besides that, here are a few thoughts from me:

> onLoadStart constructor

The name `onLoadStart` is a bit irritating (at least for me). What about onUploadStart?

> Add an onLoadStart constructor option that gets called when an upload starts.

That's not correct. In your implementation, the onLoadStart function will be invoked whenever a PATCH request starts which can happen multiple times during a single upload (e.g. if the connection is interrupted).

> This uses the XMLHttpRequest.onloadstart so will only work in browsers.

I believe it should also be able to also implement this for Node, which would help keeping this package consistent.

gavboulton:
Thanks for your comments @Acconut.

You're right, some context would definitely help. I've been looking at the [Tus10 plugin in Uppy](https://github.com/transloadit/uppy/blob/master/src/plugins/Tus10.js) and want to be able to pause an upload/all uploads using a button. I only want to show this button when I know when an upload is in progress. Uppy emits an event when an upload starts, or more specifically, [`core:upload-started` is emitted just after the upload method of tus-js-client is called](https://github.com/transloadit/uppy/blob/master/src/plugins/Tus10.js#L184-L185). At this point, in tus-js-client, the XHR for the upload hasn't been created (and isn't assigned to `this._xhr`) as there are initial POST/HEAD requests. I'd like a way to know when the main upload starts so I know that abort can be called.

You make a good point about the potential new callback option being called multiple times, so `onLoadStart` isn't right. I called it this to match the XHR event listener. Something more descriptive would be `onLoadStartContinueOrResume`, but is a 'little' verbose!

Acconut:
Thank you for the detailed description. For me this sounds more like a bug in tus-js-client and we should fix it directly in the library. I will look into this right now. What do you think?

gavboulton:
This sounds good to me. From my point of view, I'd like to be notified somehow that the 'main' part of the upload has started. Callbacks seem to be the way to go but perhaps they could be used in a different way. Another pair of eyes on this is good, thanks.

Acconut:
The latest commit should fix this issue. Would you be so kind and test if it works for your setup?

> I'd like to be notified somehow that the 'main' part of the upload has started

You still have the progress events which also indicate this, if they work for you. Generally, I like to keep the exposed interface small to make future refactorings easier. :)

gavboulton:
This looks good, thanks, it's a better idea to expose all XHRs. I'll have to setup something that uses tus-js-client directly to completely confirm but from looking I think it's fine.

gavboulton:
Closing this in favour of https://github.com/tus/tus-js-client/commit/71a9fe6694048142b764e16d1f97e0d1db763826

Acconut:
Glad to hear that it worked for you :)

PR #84, opened 2017-11-19
Add note about server-imposed chunkSize hardlimits

pauln:
Adds a note to the readme to mention that exceeding any server-imposed chunk size hardlimits will result in chunked uploads failing.

This is to help anyone who doesn't realise that they're hitting such limits.  See the following for further information / discussion:
- https://github.com/tus/tus-resumable-upload-protocol/issues/89
- https://github.com/tus/tus-resumable-upload-protocol/pull/93


Acconut:
Thank you, let's add this, until we figured a proper solution out.

PR #86, opened 2017-12-14
Added support for Cordova File

naranjamecanica:
Since a File object loaded with the Cordova file plugin is not exactly the same a one loaded with a file input some changes were needed to support them.

The slice method of a Cordova File returns a new File object which needs to be loaded with a FileReader. Since this is an async operation, I've implemented a promise handler to accommodate.

I don't know if it is something you want to support, but being able to do resumable uploads from an App is awesome!


Acconut:
I personally haven't used Cordova much, which is the reason for my lack of knowledge in that area but I would be happy to also have support for tus-js-client on this platform. Especially, if only such a small change is necessary.

All in all, I only have to two comments:
1. Firstly, our linter has found a few inconsistencies in your code (https://travis-ci.org/tus/tus-js-client/jobs/316609144#L1152-L1158). Would you please take care of them?
2. Secondly, the tus-js-client library does not use Promises internally, mostly for compatibility with older browsers and to avoid dragging a polyfill in (if not necessary). Therefore, I would like to ask you if you could add a callback to the `slice` methods, so we can also support asynchronous slicing. What do you think?

Acconut:
Nice work, @hannuniemela, we would be happy to support Cordova in tus-js-client. Would you mind opening a PR about this so we can discuss your code more in detail? 

Besides that I am going to close this PR since there sadly hasn't been a response from the author.

Issue #97, opened 2018-03-06
Delete the fingerprint on success.


mperrando:
https://github.com/tus/tus-js-client/blob/ebc5189eac35956c9f975ead26de90c896dbe360/lib/upload.js#L441

Take into account the following use case.

"After a file has been completely uploaded, it may happens that the user deletes it from the server, and then the same user decides to upload it again"

In such a case, the new file should be treated as a brand new file, with a new id, but if we don't remove the fingerprint from the storage, the client tries to upload using the same id.

Furthermore, the local storage keeps growing while the user loads files successfully!

The idea is to add something like the following, what do you think about it?

```
if (this.options.resume ) {
    // Remove stored fingerprint and corresponding endpoint,
    // a new upload of the same file must be treated as a different file
    Storage.removeItem(this._fingerprint);
}
```

Acconut:
> "After a file has been completely uploaded, it may happens that the user deletes it from the server, and then the same user decides to upload it again"
> In such a case, the new file should be treated as a brand new file, with a new id, but if we don't remove the fingerprint from the storage, the client tries to upload using the same id.

This is true but no the full story. When tus-js-client tries to resume an upload which has been deleted from the server it will remove the fingerprint from its storage and start a brand new upload (see
https://github.com/tus/tus-js-client/blob/master/lib/upload.js#L335-L361). Therefore this scenario is not really a big problem, IMO.

> the local storage keeps growing while the user loads files successfully!

I agree, this could be a problem. Even if the fingerprint per upload is small, it will accumulate over this.

> The idea is to add something like the following, what do you think about it?

I like it, would you be able to open a PR for this?

PR #98, opened 2018-03-09
Remove fingerprint from storage on success.

mperrando:
See #97.


Acconut:
Thank you very much, this is looking very good. However, since this is a breaking change, I would recommend to move this behaviour behind an option flags is enabled (maybe call it `clearFingerprintAfterSuccess`?)

mperrando:
Ok, no problema about that.

Should this option flag be enabled by default? Of course adding also a comment that from version X.X a breaking change is present and that the old behaviour can be restored by setting `clearFingerprintAfterSuccess ` to false?

What about the inverted option flag `leaveFingerprintAfterSuccess` defaulted to false?

Acconut:
> Should this option flag be enabled by default? Of course adding also a comment that from version X.X a breaking change is present and that the old behaviour can be restored by setting clearFingerprintAfterSuccess to false?

I would like to avoid a breaking change. Since we follow Semantic Versioning (https://semver.org/), a breaking change may only land in a new major release but we don't have the resources to maintain version 1.X and 2.Y at the same time. Furthermore, many people unfortunately don't read release notes, so such a subtle change could easily cause confusion. So I would be pleased if no default behaviour is changed.

> What about the inverted option flag leaveFingerprintAfterSuccess

If this would be enabled by default, I am also happy with this name. You can also name the flag removeFingerprintOnSuccess, similar to your branch name.

mperrando:
Ok. Done.

I didn't like very much the almost perfectly duplicated test, so I did some refactor introducing a new context level dedicated to the urls storing.

Acconut:
Thank you very much, this has landed in https://github.com/tus/tus-js-client/commit/926977d69067f6993699485e03884e0be88cc378. I only added a few changes for the documentation and comments

Issue #71, opened 2017-05-09
Cannot assign to read only property 'exports' of object '#<Object>' (mix require and export)


figitaki:
I'm getting this error when importing this module in a project that uses webpack and babel. I'm using the following babel config to enable webpack's tree shaking.

```js
// .babelrc
{
  "presets": [
    [
      "latest",
      {
        "es2015": {
          "modules": false
        }
      }
    ],
    "react",
    "stage-0"
  ]
}
```

This causes an error when I import the module (webpack/webpack#4039) since the `lib/index.js` mixes `import` with `module.exports`. As a workaround I simply import the `lib.es5/index` to force babel to transpile the entire source code.

```js
import tus from 'tus-js-client/lib.es5'

/* ... */
```

But a better fix would be to not mix the two module formats.

Acconut:
I am not able to reproduce your issues. Could you please also include the necessary sections from your Webpack configuration? Also, which versions of babel and webpack are you using?

Acconut:
Closing this issue due to a lack of information. Feel free to leave a comment if you want to continue the discussion :)

ChristianDavis:
I ran into same issue, code is incorrect https://github.com/tus/tus-js-client/pull/103

PR #103, opened 2018-03-19
incorrect mix import and module.exports

ChristianDavis:
It's incorrect to mix these two syntaxes. You have an es5 module, not sure why you would want to even try using modules.export in this file.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -18,10 +18,7 @@ if (typeof window !== "undefined") {
   var isSupported = true;
 }
 
-// The usage of the commonjs exporting syntax instead of the new ECMAScript
-// one is actually inteded and prevents weird behaviour if we are trying to
-// import this module in another module using Babel.
-module.exports = {
+export default {
   Upload,
   isSupported,
   canStoreURLs,


Acconut:
As the comment you removed explained, this was used to work around an issue if you required tus-js-client using Babel. Since none was able to provide a minimal example in #71 where this setup would lead to problems, this was never changed as we could not reproduce problems. Could you share one?

ChristianDavis:
I see where this has gone awry. You can test this yourself in 5 minutes, just try it with a plain create-react-app, lib.es5 file is correct:
```
import Tus from 'tus-js-client/lib.es5/';
console.log(Tus)
 > {Upload: ƒ, isSupported: true,...
```

```
import Tus from 'tus-js-client/lib/index.js';
console.log(Tus)
 15:12-15 "export 'default' (imported as 'Tus') was not found in 'tus-js-client/lib/index.js'
```

A full load of that version of the code will issue: "Uncaught TypeError: Cannot assign to read only property 'exports' of object '#<Object>'"
Because your import statement is going to cause this file to be treated as ES6, and static analysis fails because of the module.exports

That comment made no sense: "This ES5 is a work around for if you are importing using Babel." Babel is configurable consumer of this interface, this comment should not know or care how Babel is configured. This interface was broken for the particular implementation of the consumer. 

Typically you would have one version of the source in ES6, then publish a transpiled ES5 (if you wanted it to be es5 compatible), or leave it in ES6. You have one in es5, one in a mix of es5 and es6 that I'm assuming is being transpiled specifically as es5 + import, then shipping both versions. The package itself points to the lib.es5 version as a default, but your dist files uses the lib version to build. The untranspiled lib file is still being published and sometimes imported by your other products to then be transpiled again, where it may or may not fail

Acconut:
Thank you for your explanation!

> import Tus from 'tus-js-client/lib.es5/';
> import Tus from 'tus-js-client/lib/index.js';

The package.json's main field points to `lib.es5/index.js`, so when you just require tus-js-client there is no problem, isn't it? Could you explain why someone would want to circumvent this and instead require `lib/index.js` directly?

> Because your import statement is going to cause this file to be treated as ES6, and static analysis fails because of the module.exports

Ok, that makes sense. So we could solve this by not using the import syntax in `lib/index.js` and instead only use require/module.exports there?

> That comment made no sense: "This ES5 is a work around for if you are importing using Babel." Babel is a configurable consumer of this interface, this comment should not know or care how Babel is configured. This interface was broken for the particular implementation of a consumer.

I don't remember the details of this issue but we were only able to reproduce it when using Babel. That's why it landed in the comment. If I recall correct, it had something to do with the `jsnext:main` field in the package.json

> You have one in es5, one in a mix of es5 and es6 that I'm assuming is being transpiled specifically as es5 + import, then shipping both versions.

The original source code in `lib/` is written using the ES6 import syntax with the only exception being `lib/index.js` due to the (questionable) reason. This version is then transpiled into a plain ES5 version into `lib.es5/`. I don't see where we would have ES5 with import syntax elsewhere?

> The package itself points to the lib.es5 version as a default, but your dist files uses the lib version to build. 

You are right, do you see a problem in this fact?

Acconut:
Also, the Travis build for your PR is failing: https://travis-ci.org/tus/tus-js-client/jobs/355478380
Would you mind looking into this?

Acconut:
Closing this issue due to a lack of information. Feel free to leave a comment if you want to continue the discussion :)

Issue #106, opened 2018-03-25
Uploads with size 0 bytes won't work


fentas:
Using version `"tus-js-client": "^1.4.5"`

**Some information**

_tus-js-client_
```
{ chunkSize: 16384,
  endpoint: 'http://tusd:1080/files/',
  retryDelays: [ 0, 1000, 3000, 5000 ],
  metadata: 
  { name: 'ttest',
    authorization: '**',
    directoryId: '**' },
  onError: [AsyncFunction: onError],
  uploadSize: 0,
  onSuccess: [AsyncFunction: onSuccess] }
```

_tusd_
```
[tusd] event="UploadCreated" id="b646a2d6f130e9360442ffa11bc8ec28" size="0" url="http://tusd:1080/files/b646a2d6f130e9360442ffa11bc8ec28" 
[tusd] event="ResponseOutgoing" status="201" method="POST" path="" 
[tusd] event="RequestIncoming" method="HEAD" path="b646a2d6f130e9360442ffa11bc8ec28" 
[tusd] event="ResponseOutgoing" status="200" method="HEAD" path="b646a2d6f130e9360442ffa11bc8ec28" 
```
Missing `UploadFinished`, `HookInvocationStart` and `HookInvocationFinish`.

Maybe this is an issue of `tusd`?

**Edit:** tusd should be able to handle 0 byte files.
> The Upload-Length request and response header indicates the size of the entire upload in bytes. The value MUST be a non-negative integer.
https://tus.io/protocols/resumable-upload.html#upload-length

Acconut:
What exactly is not working? If only the upload-finished hook is not invoked, then this issue is entirely related to tusd. I assume that tusd is able to handle empty uploads but the upload-finished hook is not invoked by a missing check.

fentas:
Hm. Firstly the `onSuccess` is not fired and nothing is happening.
I guess tusd is waiting for a `PATCH` request but only got a `POST`.

I am not sure if tus specs require a `PATCH` request to finish an upload?

Acconut:
> I am not sure if tus specs require a PATCH request to finish an upload?

No, the specification does not need a PATCH request. If you have an empty upload, it's enough to just send the POST request.

> I guess tusd is waiting for a PATCH request but only got a HEAD/POST.

There was a missing check in tusd, which did not fire the post-finish hook for empty uploads but that is fixed in https://github.com/tus/tusd/commit/c470fe3a9dfbc40b74aed9518138993df2fc3240.

[...]

PR #107, opened 2018-04-03
No need to start upload if filesize is 0

fentas:
fixes https://github.com/tus/tus-js-client/issues/106

This assumes that if `upload_size` equals `0` it will finish the upload after the file was successfully created (server side). No additional `PATCH` request will be fired.


Acconut:
Thank you very much, I merged this and added a test for it

Issue #96, opened 2018-02-26
Does this work with React Native


acollazomayer:
Im building an app with react native and i want to know if this library works with react native

kvz:
While I see no reason why it shouldn't, we haven't tested this yet and we'd be grateful if you could share your results with us!

tong233:
In fact, react native does not support the File API and Blob API, maybe react-native-fetch-blob can help you https://github.com/wkh237/react-native-fetch-blob/wiki/Web-API-Polyfills-(experimental)#user-content-blob

Acconut:
A (now deleted) comment suggested that tus-js-client is not compatible with React Native because latter one lacks support for the Blob API, which is to my limited knowledge not true anymore: https://forums.expo.io/t/react-native-now-has-full-blob-support/7652

Therefore, I assume that tus-js-client is compatible until we are proven the other way :) Feel free to comment if you have a different experience.

michaelost:
I receive document is not defined error. this package is build by browserify ( a tool for using nodejs modules in browser env) react-native is not a browser.
I think it could be used if it was built by some other tool which will port it to react-native

Acconut:
> I receive document is not defined error.

This error is unexpected, since tus-js-client never accesses the document object on its own. Would you mind sharing more details about this error, for example a stack trace?

> I think it could be used if it was built by some other tool which will port it to react-native

I don't have much knowledge in the Reactive Native topic, so what are the other tools using which it would work?

kvz:
We're trying to get tus-js-client to work with React Native now in https://github.com/tus/tus-js-client/pull/114#issuecomment-403524611. Just sharing in case it's still relevant to someone here.

arturi:
I think the issue is here: https://github.com/tus/tus-js-client/blob/e7a07af900c058a68fdad04001a9cde21a29d0c7/lib/browser/request.js#L2 

And `resolve-url` is the one calling `document`: 

https://github.com/lydell/resolve-url/blob/c99e0ad30108ccdf73e10eddbe7c8b2080c57740/resolve-url.js#L21-L22

![img_5746](https://user-images.githubusercontent.com/1199054/43696871-21321452-990e-11e8-96b0-09c58392cc44.PNG)

Acconut:
> And resolve-url is the one calling document:

[...]

PR #114, opened 2018-07-09
create .babelrc

taoqf:


kvz:
Hi, what is the advantage of this approach over having it in the package.json?

arturi:
Yeah, we are getting this in React Native build:

img_5330

And the solution would likely be to 1. move the config to `.babelrc` like @taoqf did here; and 2. add it to [`.npmignore`](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package) so build tools are not confused with babel configs they shouldn’t care for.

Are you ok with this @Acconut? Or maybe we should use [`files`](https://docs.npmjs.com/files/package.json#files).

Acconut:
Frankly, I am unsure how the Babel configuration is connected to React Native but if that fixes it, I am open for this and following PRs.

arturi:
React Native uses a build setup, which looks into babel configs for all dependencies in `node_modules`.

Renee:
> react native runs babel on deps i guess

taoqf:
I am not using react , just webpack, also come into this issue. @arturi

Acconut:
Ok, I went ahead and merged this. Let me know if anything else is needed for Webpack or React Native.

kvz:
I think what we need next is add `.babelrc` to `.npmignore` (blacklist) OR: add all files we want present in the npm module to `files` in the `package.json` (whitelist).

Personally, I'm a fan of the whitelist approach, for the simple reason that it's harder to accidentally add things to the npm registry (like secrets, or what have you). It's hard to undo a publish, and unlike git, you don't really get to explicitly commit which things go into a module (unless you use `files`! :)

Acconut:
In https://github.com/tus/tus-js-client/pull/119 @arturi added the `files` property to `package.json`.

Further work for fully supporting React Native is discussed in #96

PR #119, opened 2018-08-02
Add `files` property to only include the necessary files in the package

arturi:
Otherwise things like `.babelrc` get added, and that causes issues for React Native and others, see https://github.com/tus/tus-js-client/pull/114


Acconut:
Thanks, Artur :+1:

PR #125, opened 2018-09-07
Make connection reuse possible with TUS Client. 

michau-krakow:
I have updated Readme.md with details, so in short, this pull request make it possible to use user-provided Agent instance to issue HTTP requests.
The change makes it possible (and I use it extensively) to upload series of files, one after another, to the same server, and to reuse the same TCP connection for that.

Note: Initially I tried just to enable 'keepAlive' setting in global Agent (yes, it affects all HTTP sessions but it was acceptable for me) but this does not work as I would expect - it held connection stalled instead of reusing it.
After some digging I found https://github.com/nodejs/node/issues/9595, suggesting the HTTP request has to be properly finished before it can be reused. This change is primarily to port the suggested solution to tus-js-client.


README.md
```diff
@@ -207,6 +207,10 @@ This will only be used if the size cannot be automatically calculated. This
 is currently only used and required if you supply a `Readable` stream as the
 file to upload. You may also use this to limit the position until which a file
 will be uploaded.
+* `agent = undefined`: specifies the [Agent](https://nodejs.org/api/http.html#http_class_http_agent)
```
Acconut: We usually use `null` as a default value for optional parameters. Please use it and also update the code.

Acconut: We usually use `null` as a default value for optional parameters. Please use it and also update the code.

README.md
```diff
@@ -207,6 +207,10 @@ This will only be used if the size cannot be automatically calculated. This
...
+instance to be used to manage HTTP requests. See nodejs documentation for details.
```
Acconut: In this document we use Node.js instead of nodejs to match how they spell it on their own. Besides that, it would be nice to make it clear that this flag is only possible for the Node.js environment, and not when used inside browsers.

README.md
```diff
@@ -293,6 +297,50 @@ The underlying implementation is rather straightforward: Any error which would u
 
 If all of these conditions are met, an attempt will be issued after applying the defined delay. Furthermore, once the client was able to successfully transfer chunks of the upload to the server, the counter for attempted retries will be reset to zero. For example, if an upload is interrupted the first delay will be applied. After reconnecting to the remote endpoint, it is able to transfer data to it until the connection is cut again. This time not the second delay will be used but the first one again because we were able to upload chunks. The reason for this behavior is that it will allow uploads to be interrupted more often than the `retryDelays` option defines, as long as we are making progress in uploading.
 
+### Persistent connections
```
Acconut: Nice explanation :+1:

lib/node/request.js
```diff
@@ -32,22 +33,36 @@ class Request {
...
+    }
+
     let req = this._request = (options.protocol !== "https:")? http.request(options) : https.request(options);
     req.on("response", (res) => {
       this.status = res.statusCode;
       this._resHeaders = res.headers;
 
       this.onload();
+
+      res.resume();   // this allows Agent to reuse same connection for next HTTP request
```
Acconut: Please link to https://github.com/nodejs/node/issues/9595 in order to explain this statement.

lib/upload.js
```diff
@@ -418,6 +419,9 @@ class Upload {
 
     let xhr = newRequest();
 
+    if (this.options.agent)
+      xhr.setAgent(this.options.agent);
```
Acconut: Please use curly braces to surround the if-statement's body.

Acconut: Also, please add a check that the `setAgent` method actually exists, which is not the case for browser environments. If it's not available, an error should be thrown.

README.md
michau-krakow: I accept your coding standard, although at this point I was trying to follow https://nodejs.org/dist/latest-v6.x/docs/api/http.html#http_http_request_options_callback convention, they use 'false', 'undefined', and Agent object, and 'null' is not an option.
I'm OK for having 'null' as default option, but it would probably need special handling in the code to pass 'undefined' to HTTP request. Please comment what you prefer.

README.md
michau-krakow: Will do.

lib/node/request.js
michau-krakow: Adding link in code comment - not sure this is what you meant.

README.md
michau-krakow: Done.

README.md
michau-krakow: Done.

Acconut:
Closing this issue due to inactivity. In particular since this PR will be incompatible with the upcoming v2 release of tus-js-client. Feel free to leave a comment if you want to continue the discussion :)

PR #126, opened 2018-09-13
upload streams in browser using upload-defered-length

### LINKED ISSUE #118: Upload streams from the browser
In issue #59, @Acconut called out that it doesn't makes sense for tus-js-client to support the creation-defer-length extension until tusd supports it. However, since then, [tusd has added support](https://github.com/tus/tusd/pull/182). Does it makes sense to add this feature now?

In terms of API level changes, I imagine the `file` argument to `tus.Upload` would be expanded to allow `ReadableStream` in addition to `File` and `Blob`.

If the server doesn't allow deferred length uploads, we could either error when passed a `ReadableStream` or just read the whole stream into a `Blob` and uploader it. I don't have strong feeling either way.

I'd be happy to take a stab at implementing this if y'all think support should be added.

paulrosenzweig:
Resolves #118 

This PR adds the ability to upload streams in the browser. In order to upload data as it's created, I used the `Upload-Defer-Length` header with the creation-defer-length tus extension.

To show a possible application of this, I added a button to the demo page that captures video with your webcam and uploads the data as it is recorded.

## What kind of "streams"?

Ideally, we'd use [ReadableStreams](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream), but sadly that's not finalized yet. Instead you can upload any object adhering to a simple contract:

You can upload a `reader` like you would a file:

`new tus.Upload(reader)`

The `reader` must has a method `read` that returns a promise. The promise should resolve to an object `{value, done}`.

`value` is data as an array, blob, or typed array.
`done` is a boolean.

The `read` method described here is exactly what's (more eloquently) described in MDN [here](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/read).

`ReadableStreams` as implemented do have a `getReader` method that returns a compatible object. If you don't have access to a reader, but instead have a callback that is run as new data is available, it's straightforward to wrap this in a reader.

## What happens if `chunkSize` is not set?

We throw an error. With files, it makes sense to upload the entire file as default behavior. However, with streams, we need to know how much data to wait for before sending each chunk.

## What happens if `read` produces data small or larger than `chunkSize`?

If `read` results in more data than we want to send in one chunk, we cut off `chunkSize` bytes and save the rest for future sends.

If `read` results in less data than we want to send in one chunk, we immediately call read again until we have enough data for one chunk or the stream ends. 

## Is this "resumable"?

It's not. I'm not quite sure what it would mean to resume uploading a stream. Would we assume the stream's current read position is the same offset as the upload?

I expect the right solution to network failures is for users to tee streams to a file and if the stream upload fails they can retry with the file, but I'm very open to different options here.


Acconut:
Thank you very much for this amazing PR, @paulrosenzweig! Unfortunately, I didn't have the time to look into this yet but I will definitely try to do so in the next few days.

kvz:
Without having looked at the code yet, regarding making stream uploads resumable, we be able to buffer chunks locally and retry uploading those, up until a point? (configurable, likely not hundreds of MB’s) That would at least make the uploading resumable in the face of switching cell towers/WiFi access point/server restarts/load balancer changes/etc. or maybe I’m thinking about this in the wrong way?

Acconut:
> we be able to buffer chunks locally and retry uploading those, up until a point? (configurable, likely not hundreds of MB’s) That would at least make the uploading resumable in the face of switching cell towers/WiFi access point/server restarts/load balancer changes/etc. or maybe I’m thinking about this in the wrong way?

Yes, that should be possible. From looking at the code, the stream is already read into a buffer before being uploaded. This is also how tus-java-client handles retries when uploading a chunk fails.

paulrosenzweig:
Sorry for the delay in addressing your feedback. I finally got a chance to make this change.

I added the ability to retry the last chunk. I had been clearing data from the buffer as soon as it was read, but now I only clear data in excess of the most recent `chunkSize` bytes. If an upload fails, the source has enough data to return the most recent chunk from the buffer again.

Mostly this just required tweaking the logic around buffer offsets, but I also had to make another change related to the source. Previously, retrying an upload would recreate the source, but with the buffer state living _in_ the source, I updated this to reuse the existing source instance rather than recreate it from the file. Everything seems to still work, but let me know if you anticipate any issues that this might cause. Here's the change:

```
-    let source = this._source = getSource(file, this.options.chunkSize);
+    this._source = this._source || getSource(file, this.options.chunkSize);
```

Note: Removing the local variable `source` wasn't strictly necessary, but it was only used twice so I just referenced `this._source` in both places.

paulrosenzweig:
Hey @Acconut just wanted to bump this!

I totally understand that open source is not your full-time job, but I'd love to get this merged after any feedback.

Acconut:
Sorry for the delay, @paulrosenzweig. I should have more time now. First of all, I would like to discuss a few more high level questions:

1. I would prefer it if the `_readUntilEnoughDataOrDone` method would not use Promises. I know they may be easier to use but tus-js-client uses callbacks for its internals and the exposed interface and i would like to keep the library code consistent.

2. I am not sure about this point yet but I think we should make length-deferring opt in instead of automatically assuming that the user wants to use length-deferring if no size is specified. This makes our internals easier and we can also keep the safety check that you removed in your PR (https://github.com/tus/tus-js-client/pull/126/files#diff-26d16a76eb2641c3108d9b4b67f9cb72L96). This is just here to make spotting errors easier and avoid misconfiguration for the user.

3. What would be a good way to dealing with the upload progress callbacks? Currently we provide the users with bytesSent and bytesTotal for each progress event but we do not have bytesTotal. Does your code account for that yet? I was thinking about setting bytesTotal to null to indicate that we do not have it yet.

Let me know what you think.

paulrosenzweig:
Thanks for your thoughts!

1. This is tricky because the API I've defined above relies on a [`read`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamDefaultReader/read) method that returns a Promise. That said, I totally appreciate the value of having consistent internal style. Do you think it makes sense to restructure things internally to use callbacks wherever possible but to still use Promises to receive data from the user? Or, would you think the API should change to not accept a "reader"?

2. That makes sense to me. I'll add an option for this.

3. Yeah I wasn't sure what to do here either, but I left it as passing `null` like you suggest. I had also considered passing the current `bytesSent` as for `bytesTotal`, but thought it'd be more confusing since you'd always be at 100% progress.

Acconut:
> Do you think it makes sense to restructure things internally to use callbacks wherever possible but to still use Promises to receive data from the user?

Yes, I think that would be the best option for our current situation.

> Yeah I wasn't sure what to do here either, but I left it as passing null like you suggest.

Perfect, we will add this to the documentation and then it should be clear.

lib/browser/source.js
```diff
@@ -11,7 +11,95 @@ class FileSource {
...
+    return this._readUntilEnoughDataOrDone(start, end, callback);
+  }
+
+  _readUntilEnoughDataOrDone(start, end, callback) {
+    const hasEnoughData = end <= this._bufferOffset + len(this._buffer);
+    if (this._done || hasEnoughData) {
+      callback(this._getDataFromBuffer(start, end));
+      return;
+    }
+    this._reader.read().then(({ value, done }) => {
```
Acconut: Could you please add error handling in the case that this promise gets rejected?

lib/browser/source.js
```diff
@@ -11,7 +11,95 @@ class FileSource {
...
+    );
+    const chunkSize = end - start;
+    if (len(this._buffer) >= 2 * chunkSize) {
+      this._buffer = this._buffer.slice(chunkSize);
+      this._bufferOffset += chunkSize;
+    }
+    return bufferToReturn;
+  }
+
+  close() {}
```
Acconut: Is it possible to cancel the stream or close the reader in here?

lib/browser/source.js
```diff
@@ -20,5 +108,9 @@ export function getSource(input) {
     return new FileSource(input);
   }
 
+  if (typeof input.read === "function") {
+    return new StreamSource(input, chunkSize);
+  }
+
   throw new Error("source object may only be an instance of File or Blob in this environment");
```
Acconut: Would you mind updating this error to mention that you can also put in a ReadableStream?

lib/browser/source.js
```diff
@@ -11,7 +11,95 @@ class FileSource {
...
+    const hasAllDataBeenRead = start >= this._bufferOffset + len(this._buffer);
+    if (this._done && hasAllDataBeenRead) {
+      return null;
+    }
+    const bufferToReturn = this._buffer.slice(
+      start - this._bufferOffset,
+      end - this._bufferOffset
+    );
+    const chunkSize = end - start;
+    if (len(this._buffer) >= 2 * chunkSize) {
```
Acconut: Would you mind explaining this condition? Maybe that's also worth to be added as comment.

lib/browser/source.js
```diff
@@ -11,7 +11,95 @@ class FileSource {
...
+  This function helps StreamSource accumulate data to reach chunkSize.
+*/
+function concat(a, b) {
+  if (a.concat) {
+    return a.concat(b);
+  }
+  if (a instanceof Blob) {
+    return new Blob([a,b], {type: a.type});
+  }
+  if (a.set) {
```
Acconut: For which data type is this condition? Could you add a comment?

test/spec/upload.browser.js
```diff
@@ -171,5 +173,510 @@ describe("tus", function () {
...
+            return Promise.resolve({ value, done: false });
+          } else {
+            return Promise.resolve({ value: undefined, done: true });
+          }
+        }
+      };
+
+      var options = {
+        endpoint: "http://tus.io/uploads",
+        headers: {
```
Acconut: It's not necessary to assert custom headers, metadata and withCredentials in every test case. This just makes it harder to read. Would you mind removing them from your new test cases?

lib/browser/source.js
paulrosenzweig: This is for typed arrays. I added a comment.

test/spec/upload.browser.js
paulrosenzweig: You're totally right. I removed a lot of clutter!

lib/browser/source.js
paulrosenzweig: Embarrassingly, this code was wrong! Attempting to write a comment made me realize my error 😬

The thinking was that we can't remove code from the buffer until we know it has been successfully uploaded. I tried to achieve this by always keeping twice the chunkSize to ensure we had the last chunk of data for resending. This approach is wrong.

The size of the buffer doesn't matter, and instead we can only delete data that is _before_ the currently requested chunk. e.g. If the uploader is requesting bytes 1000-1100, we know that bytes < 1000 have been successfully uploaded.

I fixed the code, added a test, and described the behavior in a comment.

lib/browser/source.js
```diff
@@ -11,7 +11,102 @@ class FileSource {
...
+        this._done = true;
+      } else if (this._buffer === undefined){
+        this._buffer = value;
+      } else {
+        this._buffer = concat(this._buffer, value);
+      }
+
+      this._readUntilEnoughDataOrDone(start, end, callback);
+    }).catch(err => {
+      throw new Error("Error during read: "+err);
```
Acconut: You should not use `throw` in there. Since this function may be invoked asynchronously, the error would not be catchable. Instead, the error should be passed up to the callback where it is then handled by `Upload#_emitError`. This also applies to the error in `StreamSource#slice`.

paulrosenzweig: Fixed this up and also made my callback argument style match the other callbacks. Let me know how it looks!

lib/upload.js
```diff
@@ -81,41 +82,41 @@ class Upload {
       return;
     }
 
-    getSource(file, this.options.chunkSize, (err, source) => {
-      if (err) {
-        this._emitError(err);
-        return;
-      }
+    if (this._source) {
+      this._start(this._source);
```
Acconut: Would you mind explaining the rational behind this if branch? It looks a bit suspicious to me. Is it about retries?

paulrosenzweig: Yes this is needed for retries. Previously this was written as something like `this._source ||= getSource(...`. Now that `getSource` is async, it's a little more verbose.

When a retry occurs `.start()` is called again. The uploader instance has the current offset and will request a chunk at that point from the source. However, if the source is a newly instantiated `StreamSource`, it doesn't know the current offset. We need to reuse the source so it saves any bytes in its buffer and knows the byte offset at the start of that buffer.

Acconut: Ok, that makes sense. Thank you

Acconut:
Thank you very, very much for your continuous support and dealing will all of my feedback. I just added a few final touches (a bit more documentation and splitting the demo to improve code and UI usability). A new release should come in the next days with this feature.

PR #129, opened 2018-10-18
Use module instead of jsnext:main

jeetiss:
jsnext:main has been superseded by pkg.module, which indicates the location of a file with import/export declarations.

https://github.com/jsforum/jsforum/issues/5

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "1.5.1",
   "description": "A pure JavaScript client for the tus resumable upload protocol",
   "main": "lib.es5/index.js",
-  "jsnext:main": "lib/index.js",
+  "module": "lib/index.js",
   "files": [
     "lib/**/*",
     "lib.es5/**/*",


Acconut:
Thanks for bringing this up, but from what I have read about `pkg.module`, it's not that easy. The files in `lib/` (that you reference in `pkg.module`) do use classes and other newer JS features. From reading https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features, this would not be allowed.

Also, which tools and versions support `pkg.module`? Is there any information available about this?

goto-bus-stop:
`jsnext:main` is no longer in use by most bundlers.

For `module`, opinions differ on whether it should be compiled down to ES5. It absolutely must only contain standard JS though, no experimental features. To make sure that users don't have to jump through hoops, it's best to aim for ES5, IMO.

Acconut:
@goto-bus-stop Thanks for the insights, I am a bit lost in the world of JavaScript bundlers :grin: . So would you recommend following setup?
- `main` points to the transpiled ES5 code with JS module syntax turned into CommonJS
- `pkg.module` points to the transpiled ES5 code with full JS module syntax

Or is it not necessary anymore that `main` points to strictly ES5?

goto-bus-stop:
`main` can theoretically point to any valid standard JS file, but for packages for browser consumption it's still best if it's ES5.

Acconut:
This will be implemented in the v2 release.

PR #131, opened 2018-11-15
Add support for React Native 

### LINKED ISSUE #96: Does this work with React Native

arturi:
This PR adds support for React Native environment (both Expo and regular React Native). 

When in React Native environment, and a file object with a `uri` property is passed to `tus.Upload`, instead of a `Blob` that tus normally expects, we first fetch the blob from a local uri using XMLHttpRequest.

Hopefully fixes #96.

This is a draft PR, I realize things might need to be moved around or maybe reworked entirely. But it seems to be working fine locally for me, when passing `tus.Upload({ url: 'file://...' })`.


lib/upload.js
```diff
@@ -81,6 +83,27 @@ class Upload {
...
     }
 
+    // In React Native, when user selects a file, instead of a File or Blob, 
+    // you usually get a file object {} with a uri property that contains
+    // a local path to the file. We use XMLHttpRequest to fetch 
+    // the file blob, before uploading with tus.
+    if (isReactNative && 
+      typeof file === 'object' && 
+      file !== null) {
+        console.log('IT’S FUCKING WORKING!!!!!!')
```
goto-bus-stop: lets keep this :wink:

arturi: Been rapidly learning how to undo this for the past 30 minutes.

PR #132, opened 2018-11-15
Add support for React Native

### LINKED ISSUE #96: Does this work with React Native

arturi:
This PR adds support for React Native environment (both Expo and regular React Native). 

When in React Native environment, and a file object with a `uri` property is passed to `tus.Upload`, instead of a `Blob` that tus normally expects, we first fetch the blob from a local uri using XMLHttpRequest.

Hopefully fixes #96.

This is a draft PR, I realize things might need to be moved around or maybe reworked entirely. But it seems to be working fine locally for me, when passing `tus.Upload({ url: 'file://...' })`.


lib/upload.js
```diff
@@ -81,6 +83,26 @@ class Upload {
       return;
     }
 
+    // In React Native, when user selects a file, instead of a File or Blob, 
+    // you usually get a file object {} with a uri property that contains
+    // a local path to the file. We use XMLHttpRequest to fetch 
+    // the file blob, before uploading with tus.
+    if (isReactNative && 
+      typeof file === 'object' && 
+      file !== null) {
```
Acconut: Could it be that you meant to write `file.uri !== null`?

Acconut:
> On a more important note, integration tests are failing, I wonder if it’s because they don’t expect async getSource?

It's not properly asynchronous even if we use a callback there since the callback is invoked synchronously. However, that's not the problem. The issue was that `getSource` didn't pass the result to the callback but instead returned the value.

I push a fix to your branch which solves this and also clean ups a few different problems that the new callback would introduce.

arturi:
Added “check for undefined instead of null”, because it was failing in React Native when actual blob was passed 😱. Not sure this is the 💯best way to make sure we are dealing with an object, which has a `uri` property, but seems ok?

test/spec/upload.browser.js
```diff
@@ -171,5 +171,104 @@ describe("tus", function () {
...
+        upload.start();
+
+        var req = jasmine.Ajax.requests.mostRecent();
+        expect(req.url).toBe("file:///my/file.dat");
+        expect(req.method).toBe("GET");
+        expect(req.responseType).toBe("blob");
+
+        req.responseError();
+
+        expect(options.onError).toHaveBeenCalledWith(new Error("tus: cannot fetch `file.uri` as Blob, make sure the uri is correct and accessible. [object Object]"));
```
arturi: > `[object Object]`

I guess I need to remove the error text if we can’t display the actual message, `[object Object]` is not very useful. Maybe it can be converted to a string?..

PR #140, opened 2019-01-13
Extend Defer length for nodejs environment

ifedapoolarewaju:
This PR extends the defer-length implementation so it can work in the nodejs environment

### DIFF
--- a/lib/node/source.js
+++ b/lib/node/source.js
@@ -58,10 +58,18 @@ class StreamSource {
     this.size = null;
 
     stream.pause();
+    this._done = false;
+    stream.on("end", () => this._done = true);
 
     this._buf = new Buffer(chunkSize);
     this._bufPos = null;
     this._bufLen = 0;
+  }
+
+  _copyBuf(start, end) {
+    const buf = this._buf.slice(start || 0, end || this._bufLen);
+    buf.size = buf.length;
+    return buf;
   }
 
   slice(start, end, callback) {
@@ -70,9 +78,7 @@ class StreamSource {
     if (start >= this._bufPos && start < (this._bufPos + this._bufLen)) {
       let bufStart = start - this._bufPos;
       let bufEnd = Math.min(this._bufLen, end - this._bufPos);
-      let buf = this._buf.slice(bufStart, bufEnd);
-      buf.size = buf.length;
-      callback(null, buf);
+      callback(null, this._copyBuf(bufStart, bufEnd));
       return;
     }
 
@@ -83,22 +89,25 @@ class StreamSource {
       return;
     }
 
-    this._bufPos = start;
-    this._bufLen = 0;
+    if (this._done) {
+      callback(null, null, this._done);
+      return;
+    }
 
-    let bytesToSkip = start - this._bufPos;
+    let bytesToSkip = start - (this._bufPos + this._bufLen);
+    this._bufLen = 0;
+    this._bufPos = start;
     let bytesToRead = end - start;
     let slicingStream = new SlicingStream(bytesToSkip, bytesToRead, this);
     this._stream.pipe(slicingStream);
-    slicingStream.size = bytesToRead;
+
     callback(null, slicingStream);
   }
 
   close() {
-    //this._stream.
+    // not implemented
   }
 }
-
 
 class SlicingStream extends Transform {
   constructor(bytesToSkip, bytesToRead, source) {
--- a/lib/node/source.js
+++ b/lib/node/source.js
@@ -66,19 +66,16 @@ class StreamSource {
     this._bufLen = 0;
   }
 
-  _copyBuf(start, end) {
-    const buf = this._buf.slice(start || 0, end || this._bufLen);
-    buf.size = buf.length;
-    return buf;
-  }
-
   slice(start, end, callback) {
     // Always attempt to drain the buffer first, even if this means that we
     // return less data, then the caller requested.
     if (start >= this._bufPos && start < (this._bufPos + this._bufLen)) {
       let bufStart = start - this._bufPos;
       let bufEnd = Math.min(this._bufLen, end - this._bufPos);
-      callback(null, this._copyBuf(bufStart, bufEnd));
+      let buf = this._buf.slice(bufStart, bufEnd);
+      buf.size = buf.length;
+
+      callback(null, buf);
       return;
     }
 
--- a/lib/node/source.js
+++ b/lib/node/source.js
@@ -72,7 +72,6 @@ class StreamSource {
       let bufEnd = Math.min(this._bufLen, end - this._bufPos);
       let buf = this._buf.slice(bufStart, bufEnd);
       buf.size = buf.length;
-
       callback(null, buf);
       return;
     }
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -41,13 +41,13 @@
   "homepage": "https://github.com/tus/tus-js-client",
   "devDependencies": {
     "axios": "^0.18.0",
-    "babel-cli": "^6.6.5",
-    "babel-preset-es2015": "^6.1.18",
+    "babel-cli": "^6.26.0",
```
ifedapoolarewaju: there were a lot of vulnerabilities after installation so I ran `npm audit`

lib/upload.js
```diff
@@ -219,6 +219,14 @@ class Upload {
     }
   }
 
+  setSize(size) {
```
Acconut: What do we need the setSize method for? We know the size of the stream once the stream has ended. Or is there a situation where this is not enough?

ifedapoolarewaju:
A situation where it’s regular ReadStream or Readable stream from nodejs we can’t always expect this behavior for sure. For example in companion we are writing(downloading) in parallel as we are reading(with tus-js), and the reader might hit “null” if a new chunk has not been written since the last chunk was read. But it doesn’t mean that writing has ended. But we can setSize manually in this case when we know for sure that we are done writing.

lib/upload.js
ifedapoolarewaju: @Acconut  A situation where it’s regular ReadStream or Readable stream from nodejs we can’t always expect this behavior for sure. For example in companion we are writing(downloading) in parallel as we are reading(with tus-js), and the reader might hit “null” if a new chunk has not been written since the last chunk was read. But it doesn’t mean that writing has ended. But we can setSize manually in this case when we know for sure that we are done writing.

Acconut: For me this sounds like your implementation has a flaw. A stream not having data to read (i.e. returning null for a read call) and a stream being ended are two different things. That's what the "end" event is for. I have never heard that you can assume a stream is finished just because there is at the moment no data available to consumption. Could it be that there is a misunderstanding by one of us?

ifedapoolarewaju: Oh, I see what you mean. I thought you were referring to the implementation from [here](https://github.com/tus/tus-js-client/pull/126). So you are suggesting that I add the `.end` event to the read stream and remove the `setSize` method? Yeah I can do that

Acconut: The streams from #126 don't have an "end" event but their `read` method returns two values: The actual chunk (which may be null) and a boolean `done` indicating if the stream has ended (see https://github.com/tus/tus-js-client/pull/126/files#diff-94df636fe83ad802ecbd0ade3cd91ff5R42). So you don't listen for the end event but read until the done value turns true.

So, I am not sure what you mean with
>  suggesting that I add the .end event to the read stream

but if you are able to trigger [this code](https://github.com/tus/tus-js-client/pull/126/files#diff-26d16a76eb2641c3108d9b4b67f9cb72R528) when the stream ends and remove the setSize method, I would be happier :)

ifedapoolarewaju: @Acconut I have updated the PR to auto-detect the file size when it is done streaming.

ifedapoolarewaju:
Hi @Acconut this PR is blocking this fix on uppy https://github.com/transloadit/uppy/pull/1274 . Would be nice if you could review this as soon as you can 😁

Acconut:
Apologies for the delay! I haven't had a thorough look but here a two comments for now:

> I changed from the use of [Tranforms](https://github.com/tus/tus-js-client/blob/master/lib/node/source.js#L92) to the use of direct streaming because we will be wrongly setting this stream size [here](https://github.com/tus/tus-js-client/blob/master/lib/node/source.js#L94) in the case that the value of `bytesToRead` exceeds the bytesAvailable (This is only likely when `deferredLength` is enabled).

AFAIK, the `size` property is optional in the Node.js context and does not have to be present: https://github.com/tus/tus-js-client/blob/master/lib/node/request.js#L44

> I also fixed some bugs that I spotted, e.g [bytesToSkip will always be 0](https://github.com/tus/tus-js-client/blob/master/lib/node/source.js#L87-L90)

Good catch, thanks for pointing this out! This somehow feels like a bug or a feature that never worked from the beginning. I would like to investigate what I wanted to achieve with it.

ifedapoolarewaju:
> AFAIK, the `size` property is optional in the Node.js context and does not have to be present

I wasn't sure of the value of its presence since I saw that other sources implemented it.
Would you prefer that I revert to the use of `Transforms` but then exclude the `size` property? Also what is the benefit of setting the property at all?

lib/node/source.js
```diff
@@ -58,22 +58,27 @@ class StreamSource {
...
     stream.pause();
+    this._done = false;
+    stream.on("end", () => this._done = true);
 
     this._buf = new Buffer(chunkSize);
     this._bufPos = null;
     this._bufLen = 0;
   }
 
+  _copyBuf(start, end) {
```
Acconut: I wouldn't call this function _copyBuf because Buffer#slice does not create a copy of the buffer's data. Quoting the relevant section from the docs (https://nodejs.org/dist/latest-v10.x/docs/api/buffer.html#buffer_buf_slice_start_end):

> Returns a new Buffer that references the same memory as the original, but offset and cropped by the start and end indices.
[...]
Modifying the new Buffer slice will modify the memory in the original Buffer because the allocated memory of the two objects overlap.

It's not relevant for our situation as we don't modify any buffers' content. But calling this function _copyBuf even if it just creates a new reference to the same buffer data is rather confusing.

Acconut:
I had a more thorough look through your patch and there is one difference I didn't spot earlier. Let me explain:

The current code with its Transform implementation does two things in parallel when reading data: It directly pipes the data to the HTTP request for uploading and stores the data inside the internal buffer (which may be needed for retries).

With your changes, these two tasks would not be done concurrently but rather in sequence: First the data is read into the internal buffer and only if this buffer is completely filled, the entire data will be passed to the HTTP request.

Now, imagine following situation: We to download a large file from cloud provider X and upload it to the tus server Y. For that we configure tus-js-client with a chunkSize of 100MB because we want to avoid sending too many small PATCH requests. What will happen with your PR is that we first read 100MB from cloud provider X, and after that upload 100MB to tus server Y and then repeat those two steps again.

However, there is no uploading happening while we are receiving the 100MB from cloud provider X. This may not be a big deal if we have a high throughput from cloud provider X but if the connection between our machine and cloud provider X is slow, we waste some valuable time by waiting for the 100MB download to complete. The current code handles this better (but in a more complex fashion) by sending the data concurrently to the tus server Y while saving it in an internal buffer.

I think it would be better to stick to the current approach. What do you think? Or did I misunderstand your code?

ifedapoolarewaju:
@Acconut yes I think you make valid point, also if the `size` property is not a mandatory field, then I think I can revert the change. Part of the reasons I made this change is because `Transform` doesn't provide a reliable way to determine the size of the slice.

Acconut:
Yes, I do think that the `size` property *for the return value from `Source#slice()`* is not actually needed and removing it should not require many code modifications. I would be very happy if you could revert to the `Transform` approach.

ifedapoolarewaju:
@Acconut I have updated the PR to revert the change in Transform.

Acconut:
Thanks, Ife!

PR #152, opened 2019-05-06
Fingerprint for node +  combine lock error with storage error

ifedapoolarewaju:
This PR addresses issues raised in this PR https://github.com/tus/tus-js-client/pull/73

### DIFF
--- a/test/spec/upload.js
+++ b/test/spec/upload.js
@@ -67,6 +67,8 @@ describe("tus", function () {
     it("should upload a file", function (done) {
       var file = new Blob("hello world".split(""));
       var host = "http://files.tus.io";
+      var passedOptions;
+      var passedFile;
       var options = {
         endpoint: host + "/uploads",
         headers: {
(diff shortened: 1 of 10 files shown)


test/spec/upload.js
```diff
@@ -81,15 +83,21 @@ describe("tus", function () {
         withCredentials: true,
         urlStorage: getStorage(),
         onProgress: function () {},
-        fingerprint: function () {}
+        fingerprint: function (_file, _options, cb) {
+          passedOptions = _options;
```
Acconut: I think it should be enough to use `expect(_file).toBe(file)` and the same for `options` here. Or is there a particular reason for why you didn't do it?

ifedapoolarewaju: yeah, I just wanted it to be clear at what point the assertion is actually required.
If it is added into the fingerprint function as you suggest, it may mislead one to think that it is called at that instant. Or it even just makes one unsure when the assertion would be triggered.

ifedapoolarewaju:
Resolved conflict issue 👍

test/spec/upload.js
Acconut: I don't agree with you here but whatever. Let's not get hung up on this, we got more important stuff to do ;)

PR #159, opened 2019-06-24
implement Upload.terminate

ifedapoolarewaju:


lib/upload.js
```diff
@@ -242,6 +272,16 @@ class Upload {
       clearTimeout(this._retryTimeout);
       this._retryTimeout = null;
     }
+
+    if (shouldTerminate) {
+      Upload.terminate(this.url, this.options, (err, xhr) => {
+        if (err || !this._hasStorage()) {
```
Acconut: Can you explain what the purpose of `|| !this._hasStorage()` here is?

lib/upload.js
```diff
@@ -242,6 +272,16 @@ class Upload {
...
+          return cb(err, xhr);
+        }
+
+        this._storage.removeItem(this._fingerprint, cb);
+      });
+    }
```
Acconut: The callback should also be invoked if `shouldTerminate === false`, just to have a consistent API interface.

test/spec/upload.js
```diff
@@ -1025,7 +1025,22 @@ describe("tus", function () {
...
+          }
+
+          // delete the upload after it was completed
+          upload.abort(true, function (err) {
+            if (err) {
+              done.fail(err);
+              return;
+            }
+
+            validateUploadDeletion(upload, done);
```
Acconut: Can we have a test that verifies that the terminate method sends a DELETE request to the provided upload URL?

lib/upload.js
```diff
@@ -74,6 +74,36 @@ class Upload {
     this._offsetBeforeRetry = 0;
   }
 
+  static terminate(url, options, cb) {
```
Acconut: Please add documentation to the README for the two new public methods.

test/spec/upload.js
ifedapoolarewaju: added

lib/upload.js
ifedapoolarewaju: added

lib/upload.js
ifedapoolarewaju: fixed

lib/upload.js
ifedapoolarewaju: I have refactored the code so the intention is clearer

Issue #166, opened 2019-08-14
How to get the last patch request in onSuccess


alireza1272:
**Question**
First of all thank u to provide tus-js-client
I'm using tus-js-client in angular, and I uploaded videos through this package. But when It successfully ends I need to access to the **Last Patch** request. I need to access it's response headers because the X-download-Url of uploaded video located there.
I found nothing that give me the chance to access Last Patch which status is 204. 
I would be very glad, if anyone can help me.
Thanks.

**Setup details**
- Runtime environment: [Browser, Node.js, Angular]
- Used tus-js-client version: [`npm`]

Acconut:
Thank you for the kind words. Unfortunately, there is no feature to access the HTTP requests and responses directly. A person in another repo asked for this as well: https://github.com/transloadit/uppy/issues/1713#issuecomment-520577059 So we are considering it for the future but don't have the resources right now to work on this. As a workaround you might want to add an endpoint to your API to retrieve the download URL for a specific upload.

Acconut:
The v2 release will feature an `onAfterResponse` callback for these jobs. More details at https://github.com/tus/tus-js-client/issues/167.

Issue #167, opened 2019-08-15
Ideas for major v2.0 release


Acconut:
This issue is used to collect idea for a next major release of tus-js-client. Nothing is settled in stone, so feel free to express your ideas.

**Tasks before major release / with breaking changes**:
- [x] Remove `withCredentials` option
- [x] Remove `resume` option in favor of proper URL storage interface.
- [x] Add interfaces for `urlStorage` and `fileReader` options: https://github.com/tus/tus-js-client/commit/e8aea8d672e8d69ca2d34bf33061d69f74b64c04#diff-26d16a76eb2641c3108d9b4b67f9cb72R30
- [x] Add abstraction layer for different HTTP stacks (also allows easier mocks for testing)
- [x] Write fully new documentation
- [x] Document better approach to resuming uploads by using `Upload#start`
- [x] Enable retries by default
- [x] Add `onBeforeRequest` and `onAfterResponse` callbacks or similar to allow obtaining HTTP request/responses.
- [ ] Write blog post and release notes

**Tasks possible after major release / no breaking changes**:
- Restart upload if no progress events fired for a specific duration
- [x] Add support for splitting upload and uploading in parallel
- [x] Add more debugging output

**Dropped ideas**:
- ~~Drop UMD/AMD support and modernize compiling steps~~ (we will continue to use browserify and it provides the UMD bundle for free, so we will keep it)
- ~~Incorporate solution for large files in React Native (see https://github.com/tus/tus-js-client/issues/146)~~ We don't have enough experience with this approach and don't want to recommend this to end users without enough confidence

thenano:
Hi @Acconut thanks so much for the work on this. I had been looking around a lot for good upload solutions for uploading large video files. I tried some chunk splitting parallel upload solutions (it seems that most video APIs support this instead of TUS) but couldn't find anything that actually worked on a js web client without a lot of fiddling. This one worked as expected out of the box!
So just to add some requests to the list:
- Not sure if your onAfterResponse solves this issue, but I use 2 different endpoints for the initial post and the subsequent patch requests. This is because I want to use an authenticated session to create a signed upload link (to a different server) and then later PATCH to that link. This actually works now apart from the authentication headers. So I would need to pass different headers between the POST/PATCH requests.
- Ability to send separate chunks in parallel

Acconut:
[...]

PR #168, opened 2019-08-20
Exposing metadata generated on server back to the client.

jarey:
Hello,

Just proposing a possible solution to the following feature requests, as discussed on:
-  https://github.com/tus/tus-js-client/issues/166
and

- https://github.com/transloadit/uppy/issues/1713

This pull requests, just manage to get the possible metadata the server may have included, during the upload, and exposes it back to the client, so it can be accessible when the upload finishes.

So the information would be accesible on tus-js-client on metadata attribute of the file object.

And in uppy, of course it also would be accesible, in a simple manner just as follows, being retrocompatible and without any need of modication on Uppy code:

```
     uppy.on('upload-success', function(file, upload) {
        console.log("Additional metadata of uploaded file, set by the server: "+file.data.metadata.additionalProperty);
    });
```

Let me know if this kind of solution is valid on your opinion. I'm open to colaborate, and make any modification to the pull request, or solve the problem in another way.

Kind regards.

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -305,6 +305,7 @@ class Upload {
 
   _emitSuccess() {
     if (typeof this.options.onSuccess === "function") {
+      this.file.metadata = decodeMetadata(this._xhr.getResponseHeader('upload-metadata'));
       this.options.onSuccess();
     }
   }
@@ -623,6 +624,23 @@ function encodeMetadata(metadata) {
   }
 
   return encoded.join(",");
+}
+
+/**
+ * Decodes encoded metadata sent from the server in order to update client metadata,
+ * so client can user additional information sent from server.
+ * @param {encodedMetadata} encodedMetadata 
+ */
+function decodeMetadata(encodedMetadata){
+  var decoded={};
+  var keyValuePairs = encodedMetadata.split(",");
+
+  keyValuePairs.forEach(function(value) {
+    var keyValue = value.split(" ");
+    decoded[keyValue[0]]= Base64.decode(keyValue[1]);
+  });
+
+  return decoded;
 }
 
 /**


Acconut:
Thank you very much for your efforts, @jarey! I am not quite sure about this change yet as the tus specification does not support sending the Upload-Metadata header in responses to PATCH requests, it is only returned for HEAD requests (just to add: the specification does not prohibit it either). So I am not aware of any tus server where this would be helpful since they usually don't return Upload-Metadata headers except for HEAD requests.

jarey:
@Acconut no problem, just let me know if it will get discarted.
I have no problem as said, with changing the implementation in the manner you think is more appropiate.
I avoided to make the change managing custom headers back to the client in the response, since the changes would be bigger and the simplest solution as a POC for me was doing the changes above.
I've made my tests with the following implementation, https://github.com/tomdesair/tus-java-server just modifying the returned upload-metadata value header, when the upload is finished. 

If i'm not worng, this implementation of the server, already returned the upload-metadata in that case (but without any modifications on the data sent by the client); so the logical modification for me was taking advantage of the already implemented code and just tweak that little bit on the client.

As said just let me know and i will try to send another PR if needed with a different approach.

Acconut:
Frankly, I would prefer to not use this approach. My currently favored solution would be to introduce a `onAfterResponse`option (see https://github.com/tus/tus-js-client/issues/167) which gets invoked whenever a HTTP response comes in. The callback would then be able to inspect it for custom details, such as headers. What do you think about this?

If you want to fetch the metadata separately, you can send a `HEAD` request to the upload URL. The response should then include the Upload-Metadata header.

jarey:
@Acconut, i would give a try to the  onAfterResponse reference you mentioned and try to come up with a solution following that path. I close this PR then since will try to make a new PR with the new approach.

Kind regards.

Acconut:
> i would give a try to the onAfterResponse reference you mentioned and try to come up with a solution following that path

Sounds good! Looking forward to your solution!

PR #169, opened 2019-09-09
remove check preventing cordova from generating fingerprints

GrandUltimate:
Fixes breaking change preventing uploads from working on Cordova.

### DIFF
--- a/lib/browser/fingerprint.js
+++ b/lib/browser/fingerprint.js
@@ -1,5 +1,4 @@
 import isReactNative from "./isReactNative";
-import isCordova from "./isCordova";
 
 /**
  * Generate a fingerprint for a file which will be used the store the endpoint
@@ -9,10 +8,6 @@ import isCordova from "./isCordova";
  * @param {Function} callback
  */
 export default function fingerprint(file, options, callback) {
-  if (isCordova()) {
-    return callback(new Error("Fingerprint cannot be computed for file input type"));
-  }
-
   if (isReactNative()) {
     return callback(null, reactNativeFingerprint(file, options));
   }


Acconut:
Can you please explain what the problem was and how this change fixes it?

GrandUltimate:
The problem is being unable to upload when on Cordova with the breaking change being introduced in commit https://github.com/tus/tus-js-client/commit/99ac9964035b610cb2d7b9defccc8d3366a43684

This change fixes the the issue by removing the check to throw an error if Cordova is detected.
The fingerprint is generated just fine on Cordova. I see no reason why this condition to throw an error if on Cordova was introduced but removing it restores working functionality on Cordova.

Acconut:
Ok, now I get what you are talking about. I don't know why this check was introduced in the first place but hopefully @ifedapoolarewaju (the author of that commit) can shine some light on the situation.

ifedapoolarewaju:
what I remember vaguely is that @Acconut  mentioned something about a weird behaviour with the file system in cordova, which in turn causes problems when generating fingerprints. And so we concluded either to find how to fix the issue (which seemed not to be clear), or not to support it at all (i.e throw an error for cordova).

However, I think the conversation was over a call and it is possible I might have misunderstood what was actually being said.

GrandUltimate:
I'd be happy to help investigate and fix any potential issues related to Cordova. If you have any insight into this LMK. So far we have not had any issues using this package for our Cordova apps.

drastik:
Awesome work on this project, thank you maintainers and @GrandUltimate for the patch!  
  
When do you think you'll create a new release?  Uppy is requiring this library pinned to a release version (non-latest/master).  

Do you need any further help here?

Acconut:
> When do you think you'll create a new release?

There is concrete day yet but we are planning a new release in the next weeks. There are some other tasks we would like to address first. If that doesn't work for you, we could consider an earlier patch release.

Acconut:
@drastik v1.8.0 has been released.

PR #174, opened 2019-11-03
Retry terminate

ifedapoolarewaju:
fixes this issue https://github.com/tus/tusd/issues/320

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -90,9 +90,8 @@ class Upload {
       if (shouldRetry(err, retryAttempt, options)) {
         const delay = options.retryDelays[retryAttempt++];
         setTimeout(() => Upload.terminate(url, options, cb), delay);
-        return;
       } else {
-        return cb(err);
+        cb(err);
       }
     };
 
@@ -269,7 +268,6 @@ class Upload {
 
     cb = cb || function () {};
     if (shouldTerminate && this.url != null) {
-      this._retryAttempt = 0;
       Upload.terminate(this.url, this.options, (err) => {
         if (err) {
           return cb(err);
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -275,7 +275,7 @@ class Upload {
     }
 
     cb = cb || function () {};
-    if (shouldTerminate) {
+    if (shouldTerminate && this.url != null) {
       Upload.terminate(this.url, this.options, (err, xhr) => {
         if (err) {
           return cb(err, xhr);
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -298,7 +298,7 @@ class Upload {
 
           this._hasStorage() ? this._storage.removeItem(this._fingerprint, cb) : cb();
         });
-      }
+      };
 
       terminate();
     } else {
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -85,12 +85,24 @@ class Upload {
       options = {};
     }
 
+    let retryAttempt = 0;
+    const retryIfShould = (err) => {
+      if (shouldRetry(err, retryAttempt, options)) {
+        const delay = options.retryDelays[retryAttempt++];
+        setTimeout(() => Upload.terminate(url, options, cb), delay);
+        return;
+      } else {
+        return cb(err);
+      }
+    };
+
     let xhr = newRequest();
     xhr.open("DELETE", url, true);
 
     xhr.onload = () => {
       if (xhr.status !== 204) {
-        cb(new DetailedError(new Error("tus: unexpected response while terminating upload"), null, xhr));
+        const err = new Error("tus: unexpected response while terminating upload");
+        retryIfShould(new DetailedError(err, null, xhr));
         return;
       }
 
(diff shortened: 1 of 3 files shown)


lib/upload.js
```diff
@@ -276,13 +284,23 @@ class Upload {
...
     if (shouldTerminate && this.url != null) {
-      Upload.terminate(this.url, this.options, (err, xhr) => {
-        if (err) {
-          return cb(err, xhr);
-        }
+      const terminate = () => {
+        Upload.terminate(this.url, this.options, (err, xhr) => {
+          if (err) {
+            if (this._shouldRetry(err)) {
+              const delay = this.options.retryDelays[this._retryAttempt++];
```
Acconut: You should `this._retryAttempt = 0;` before starting the retries for the termination.

lib/upload.js
```diff
@@ -262,6 +243,33 @@ class Upload {
...
+    if (typeof window !== "undefined" &&
+        "navigator" in window &&
+        window.navigator.onLine === false) {
+      isOnline = false;
+    }
+
+    return isOnline;
+  }
+
+  _shouldRetry(err) {
```
Acconut: :+1: for refactoring. Could you add JSDoc-style comments to the new functions, in a similar fashion how (some) other methods already have?

lib/upload.js
```diff
@@ -262,6 +243,33 @@ class Upload {
...
+    let status = err.originalRequest ? err.originalRequest.status : 0;
+    let isServerError = !inStatusCategory(status, 400) || status === 409 || status === 423;
+    return this.options.retryDelays != null &&
+                      this._retryAttempt < this.options.retryDelays.length &&
+                      err.originalRequest != null &&
+                      isServerError &&
+                      this._isOnline();
+  }
+
   abort(shouldTerminate, cb) {
```
Acconut: Please also update the README to explain that the termination request is retried.

lib/upload.js
```diff
@@ -276,13 +284,23 @@ class Upload {
 
     cb = cb || function () {};
     if (shouldTerminate && this.url != null) {
-      Upload.terminate(this.url, this.options, (err, xhr) => {
-        if (err) {
-          return cb(err, xhr);
-        }
+      const terminate = () => {
+        Upload.terminate(this.url, this.options, (err, xhr) => {
```
Acconut: It would be nice if the `Upload.terminate` function would be able to retry on its own, so you can benefit from the retries without having an instance of the `tus.Upload` class. Could you look into if that's easily possible?

Acconut: The callback for `Upload.terminate` does not provide the second argument `xhr` if I am not mistaken. See https://github.com/tus/tus-js-client/blob/master/lib/upload.js#L93

lib/upload.js
ifedapoolarewaju: oh ok, I wasn't sure if it was desired or not

lib/upload.js
ifedapoolarewaju: Hmm, seems I got confused by the brackets. But do you think we should be passing it as a matter of fact? 🤔 It suddenly feels unnecessary to have it since the xhr instance is always added to the error instance. Also removing it will not be a breaking change since it was never there.

Acconut: > But do you think we should be passing it as a matter of fact?

No, the XHR is already part of the error object if available, as you also said. Please also remove.

ifedapoolarewaju:
Addressed all issues 👍

lib/upload.js
```diff
@@ -276,9 +269,10 @@ class Upload {
 
     cb = cb || function () {};
     if (shouldTerminate && this.url != null) {
-      Upload.terminate(this.url, this.options, (err, xhr) => {
+      this._retryAttempt = 0;
```
Acconut: I think setting this._retryAttempt is now unnecessary here because Upload.terminate has it's own retry counter in form of the `retryAttempt` local variable, right?

lib/upload.js
```diff
@@ -85,20 +85,32 @@ class Upload {
       options = {};
     }
 
+    let retryAttempt = 0;
+    const retryIfShould = (err) => {
+      if (shouldRetry(err, retryAttempt, options)) {
+        const delay = options.retryDelays[retryAttempt++];
+        setTimeout(() => Upload.terminate(url, options, cb), delay);
+        return;
```
Acconut: This return statement can be removed.

lib/upload.js
```diff
@@ -85,20 +85,32 @@ class Upload {
...
+      } else {
+        return cb(err);
```
Acconut: This return statement can be removed as well.

lib/upload.js
ifedapoolarewaju: resolved

lib/upload.js
ifedapoolarewaju: resolved

lib/upload.js
ifedapoolarewaju: resolved

PR #190, opened 2020-04-29
bugfix - keep streaming buffer around until upload completes

nvolungis:
**What's happening now**
When a segment fails to upload _and_ it's deemed retryable, the tus client issues a request to the server asking for the latest byte offset. That offset is then passed to the `source` as the starting position for the next segment of data. This will result in a `Requested data is before the reader's current offset` error. 

**Why is this happening**
The issue is caused by the source deleting data that has already been read. The idea was solid, but there doesn't appear to be a way to verify for sure what data is okay to delete outside of issuing an offset request to the server. Even if we had it, I don't see a clean way to get that info back into the source. 

**How this fixes it**
The solution is to keep the buffer around until the upload is complete. I think the intention was to prevent unbounded buffer growth, but it's okay to keep around for two reasons:

1 - When uploading a normal file, the entire thing is held in a buffer without issue.
2 - Albeit anecdotal, I've tested this with two simultaneous video recordings up to 20 minutes long without issue. 

If we wanted to try and delete data from the buffer periodically, my testing suggests that the retry only goes back a few segments, so we could probably safely delete data 5 or so segments back from the last read. I would like to avoid this however because it makes reading the buffer more complex and stateful.

### DIFF
--- a/lib/browser/source.js
+++ b/lib/browser/source.js
@@ -32,22 +32,16 @@ class StreamSource {
   constructor(reader, chunkSize) {
     this._chunkSize = chunkSize;
     this._buffer = undefined;
-    this._bufferOffset = 0;
     this._reader = reader;
     this._done = false;
   }
 
   slice(start, end, callback) {
-    if (start < this._bufferOffset) {
-      callback(new Error("Requested data is before the reader's current offset"));
-      return;
-    }
-
     return this._readUntilEnoughDataOrDone(start, end, callback);
   }
 
   _readUntilEnoughDataOrDone(start, end, callback) {
-    const hasEnoughData = end <= this._bufferOffset + len(this._buffer);
+    const hasEnoughData = end <= len(this._buffer);
     if (this._done || hasEnoughData) {
       var value = this._getDataFromBuffer(start, end);
       callback(null, value, value == null ? this._done : false);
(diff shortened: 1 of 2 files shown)


Acconut:
I see the problem but I disagree with the solution.

> I think the intention was to prevent unbounded buffer growth, but it's okay to keep around for two reasons

Exactly, the code has been designed to avoid excessive memory usage.

> 1 - When uploading a normal file, the entire thing is held in a buffer without issue.

Not exactly. If you select a file to upload in your browser, the entire file will not be read into memory. Instead only the file's sections that are actually uploaded will be held in memory to avoid excessive memory usage.

> 2 - Albeit anecdotal, I've tested this with two simultaneous video recordings up to 20 minutes long without issue.

What were the actual files sizes in the end? Could you also talk about what the memory consumption inside the browser were?

nvolungis:
@Acconut I can run some tests to get some data around memory usage and file sizes. I'll check out Activity Monitor (I'm on macOS) for mem usage - if you have a better idea I'd love to hear it!

> If you select a file to upload in your browser, the entire file will not be read into memory. Instead only the file's sections that are actually uploaded will be held in memory to avoid excessive memory usage.

Read up on this a bit more and realized that `file.slice` actually pulls data into memory, which I didn't know before.

I suppose that leaves us with what an acceptable approach would look like. Looking to the server for the latest offset seems like the only approach that doesn't involve guesswork. Do you have any other thoughts? 

I already tried looking at the byte ranges from each successful segment response, but those ranges don't match up with the response from the latest offset request, so even when a segment uploads completely we can't mark it as deletable. 

For what it's worth, this feels somewhat urgent to me. For reasons I still don't fully understand the chance of getting a 409 response from any single request to a tus server is relatively significant (in my experience), so for any stream of sufficient segments you approach 100% chance of the entire upload failing. This normally isn't a huge issue since retrying seems to solve the problem most of the time, but retrying is broken for streaming sources.

Acconut:
> I can run some tests to get some data around memory usage and file sizes. I'll check out Activity Monitor (I'm on macOS) for mem usage - if you have a better idea I'd love to hear it!

Most browsers (at least Firefox and Chrome) have a built in task manager to show the memory and CPU usage per browser tab. In addition, the developer consoles also offer you to monitor memory usage. I think that those two tools offer more precise measuring than using the OS activity monitor.

> Looking to the server for the latest offset seems like the only approach that doesn't involve guesswork. Do you have any other thoughts?

FWIW, the tus server always include the new offset after a successful PATCH request. So the client already has that information at hand.

> For reasons I still don't fully understand the chance of getting a 409 response from any single request to a tus server is relatively significant (in my experience), so for any stream of sufficient segments you approach 100% chance of the entire upload failing.

Frankly, this sounds like a problem with the tus server where it does not properly persist data. A 409 should very rarely happen. What tus server are you using? We have been running tusd in production for years and 409 have never been a problem. I think we should tackle the root cause at the server instead of patching the clients first.

Acconut:
Closing this issue due to inactivity. Feel free to leave a comment if you want to continue the discussion :)

PR #192, opened 2020-05-11
Remove core-js and provide transpiled ES Module output

goto-bus-stop:
This does two things 'restore' size and bundler support to v1.x levels:

## Disable `@babel/preset-env` `"useBuiltIns"`
`useBuiltIns` injects loads of polyfills that are installed into the
environment. core-js polyfills are very thorough, but that sometimes
means that it includes things that are actually totally unnecessary.

For example, tus-js-client@2 shipped with a Symbol polyfill, even though
it doesn't use Symbols anywhere. But transpiled array rest spread syntax uses
`Symbol.iterator`, so preset-env figures that `Symbol` is in use,
and adds the polyfill. This requires core-js to override things like
`Object.getOwnPropertyNames` to exclude polyfilled Symbols, etc, and it
just spirals out of control until you have 50KB of polyfills just for
array spread syntax and Promises :)

The only feature that tus-js-client _actually_ appears to require is
`Promise`, which we can reasonably expect users to polyfill (and is already
documented as a requirement).

## Transpile `package.json` `"module"`
`"module"` and `"main"` are both commonly used by bundlers. To support
the widest range of browsers, they should both be transpiled to ES5.
Now `lib.esm` contains ES5 syntax + ES Modules syntax, so bundlers with
ES Modules support can use it. For the browserify build scripts, we need
to add `@babel/plugin-transform-modules-commonjs` manually via a command
line parameter.

It's possible and not weird to use newer syntax, but best if people opt
in to it, which they can do with a resolve alias.
In webpack for example:

```js
resolve: {
  alias: {
    'tus-js-client': path.join(__dirname, 'node_modules/tus-js-client/lib')
  }
}
```

### DIFF
--- a/test/spec/helpers/assertUrlStorage.js
+++ b/test/spec/helpers/assertUrlStorage.js
@@ -8,9 +8,9 @@ module.exports = async function assertUrlStorage(urlStorage) {
   const key2 = await urlStorage.addUpload("fingerprintA", { id: 2 });
   const key3 = await urlStorage.addUpload("fingerprintB", { id: 3 });
 
-  expect(key1.startsWith("tus::fingerprintA::")).toBe(true);
-  expect(key2.startsWith("tus::fingerprintA::")).toBe(true);
-  expect(key3.startsWith("tus::fingerprintB::")).toBe(true);
+  expect(/^tus::fingerprintA::/.test(key1)).toBe(true);
+  expect(/^tus::fingerprintA::/.test(key2)).toBe(true);
+  expect(/^tus::fingerprintB::/.test(key3)).toBe(true);
 
   // Query the just stored uploads individually
   result = await urlStorage.findUploadsByFingerprint("fingerprintA");
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,8 @@
 {
 	"presets": [
-		[ "@babel/preset-env", { "useBuiltIns": "usage", "corejs": 3 } ]
+		[ "@babel/preset-env", {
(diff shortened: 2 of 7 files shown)


test/spec/helpers/assertUrlStorage.js
```diff
@@ -8,9 +8,9 @@ module.exports = async function assertUrlStorage(urlStorage) {
   const key2 = await urlStorage.addUpload("fingerprintA", { id: 2 });
   const key3 = await urlStorage.addUpload("fingerprintB", { id: 3 });
 
-  expect(key1.startsWith("tus::fingerprintA::")).toBe(true);
-  expect(key2.startsWith("tus::fingerprintA::")).toBe(true);
-  expect(key3.startsWith("tus::fingerprintB::")).toBe(true);
+  expect(/^tus::fingerprintA::/.test(key1)).toBe(true);
+  expect(/^tus::fingerprintA::/.test(key2)).toBe(true);
+  expect(/^tus::fingerprintB::/.test(key3)).toBe(true);
```
goto-bus-stop, line 13: Instead of adding a polyfill for `startsWith`, I just used an older way of doing the same thing. This should prob continue to be done for test code, else we might have more polyfills in the tests than we expect users to have, and we could accidentally depend on them in the library code.

Acconut: Absolutely agree, you are right! :+1:

Acconut:
Thank you very, very much!

Issue #197, opened 2020-05-18
Callback for deciding whether to retry or not


PVince81:
**Is your feature request related to a problem? Please describe.**
In our use case, we would like to abort directly on status codes 403, 409 and 423 as they might be legitimate errors. Also see https://github.com/tus/tus-js-client/issues/196

**Describe the solution you'd like**
Instead of hard-coding status values, provide a callback method with the response as argument to let library users decide whether to retry or not. Basically converting `shouldRetry` to a callback or parts of it.

**Describe alternatives you've considered**
Alternative solution for a library user would be to not use the retry delays and implement the retry logic on their own. But this is more cumbersome and defeats the purpose of having a retry ability in the library.

**Can you provide help with implementing this feature?**
Yes

**Additional context**

PVince81:
I'm working on a PR, stay tuned

PVince81:
PR is here: https://github.com/tus/tus-js-client/pull/198

PR #198, opened 2020-05-18
Added onShouldRetry callback for controlling retry

PVince81:
Whenever the library is about to retry an upload due to an
error, the new optional callback onShouldRetry will be
called when defined. Its return value will tell the library whether
to actually retry the upload or fail with an error, for example
based on status code checks.

This makes it possible to customize the behavior like reacting on
specific status codes.

Fixes https://github.com/tus/tus-js-client/issues/197

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,14 @@ input.addEventListener("change", function(e) {
         },
         onSuccess: function() {
             console.log("Download %s from %s", upload.file.name, upload.url)
+        },
+        onShouldRetry: function(err, retryAttempt, options) {
+          var status = err.originalResponse ? err.originalResponse.getStatus() : 0;
+          // in case the status if 403, fail directly
+          if (status === 403) {
+            return false;
+          }
+          return true;
         }
     })
 
--- a/README.md
+++ b/README.md
@@ -39,14 +39,6 @@ input.addEventListener("change", function(e) {
         },
         onSuccess: function() {
             console.log("Download %s from %s", upload.file.name, upload.url)
-        },
-        onShouldRetry: function(err, retryAttempt, options) {
-          var status = err.originalResponse ? err.originalResponse.getStatus() : 0;
-          // in case the status if 403, fail directly
-          if (status === 403) {
-            return false;
-          }
-          return true;
         }
     })
 
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -924,24 +924,19 @@ function shouldRetry(err, retryAttempt, options) {
   // - retryDelays option is set
   // - we didn't exceed the maxium number of retries, yet, and
   // - this error was caused by a request or it's response and
-  // - the error is server error (i.e. no a status 4xx or a 409 or 423) or
+  // - the error is server error (i.e. not a status 4xx except a 409 or 423) or
   // a onShouldRetry is specified and returns true
   // - the browser does not indicate that we are offline
-  let status = err.originalResponse ? err.originalResponse.getStatus() : 0;
+  if (options.retryDelays == null || retryAttempt >= options.retryDelays.length || err.originalRequest == null) {
+    return false;
+  }
 
-  // define as function so we only call it if really needed
-  let doRetry = () => {
-    if (options && typeof options.onShouldRetry === "function") {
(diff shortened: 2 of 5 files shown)


PVince81:
the tests are green, seems Travis didn't report back to Github

README.md
```diff
@@ -39,6 +39,14 @@ input.addEventListener("change", function(e) {
         },
         onSuccess: function() {
             console.log("Download %s from %s", upload.file.name, upload.url)
+        },
+        onShouldRetry: function(err, retryAttempt, options) {
```
Acconut, line 43: Please move this example to its own section in docs/usage.md. The example in the README.md is meant as the  minimum useful code, so we try to avoid adding too many options here.

lib/upload.js
```diff
@@ -923,15 +924,24 @@ function shouldRetry(err, retryAttempt, options) {
   // - retryDelays option is set
   // - we didn't exceed the maxium number of retries, yet, and
   // - this error was caused by a request or it's response and
-  // - the error is server error (i.e. no a status 4xx or a 409 or 423) and
+  // - the error is server error (i.e. no a status 4xx or a 409 or 423) or
```
Acconut, line 927: Can you please help improve this sentence? I think it should be "not a 4XX status except a 409 or 423"

lib/upload.js
```diff
@@ -923,15 +924,24 @@ function shouldRetry(err, retryAttempt, options) {
...
+    } else {
+      return !inStatusCategory(status, 400) || status === 409 || status === 423;
+    }
+  };
   return options.retryDelays != null &&
          retryAttempt < options.retryDelays.length &&
          err.originalRequest != null &&
-         isServerError &&
-         isOnline();
+         isOnline() &&
```
Acconut, line 943: If we already expose a generic function for determining retries, I think we should also remove the online check if the user supplied onShouldRetry. There are use cases where people want to overwrite that check, I believe. What do you think?

lib/upload.js
```diff
@@ -923,15 +924,24 @@ function shouldRetry(err, retryAttempt, options) {
...
   // - this error was caused by a request or it's response and
-  // - the error is server error (i.e. no a status 4xx or a 409 or 423) and
+  // - the error is server error (i.e. no a status 4xx or a 409 or 423) or
+  // a onShouldRetry is specified and returns true
   // - the browser does not indicate that we are offline
   let status = err.originalResponse ? err.originalResponse.getStatus() : 0;
-  let isServerError = !inStatusCategory(status, 400) || status === 409 || status === 423;
+
+  // define as function so we only call it if really needed
+  let doRetry = () => {
```
Acconut, line 933: Can you please refactor the code to not use an anonymous function here? I have something like this in mind:

```js
if (retryDelays == null || retryAttempt < retryDelays.lenght || originalRequest == null) {
  return false;
}

if (options && typeof options.onShouldRetry === "function") {
  return options.onShouldRetry(err, retryAttempt, options);
}

return !inStatusCategory(status, 400) || status === 409 || status === 423 && online();
```

That's easier to read IMO.

test/spec/test-common.js
```diff
@@ -837,6 +837,130 @@ describe("tus", function () {
...
+    // response of the onShouldRetry is true
+    it("should retry the upload when callback specified and returns true", async function () {
+      const testStack = new TestHttpStack();
+      var file = getBlob("hello world");
+      var options = {
+        httpStack: testStack,
+        endpoint: "http://tus.io/files/",
+        retryDelays: [10, 10, 10],
+        onSuccess: waitableFunction("onSuccess"),
+        onShouldRetry: () => true
```
Acconut, line 850: Can you add assertions that onShouldRetry is called with the correct parameters? You can do that using the spyOn functions.

PVince81:
I've addressed all comments.

Had some struggles with `spyOn()` as `callThrough()` didn't work when spying on `upload._emitError` directly to get the error object. It seems that spying on the prototype object worked.

PVince81:
@Acconut anything left to do or is this good to merge ? thanks

PVince81:
@Acconut any concerns merging this addition ?

PR #203, opened 2020-05-29
Adding options for NodeJs ClientRequest

dragoscirjan:


Acconut:
Thank you for the PR but I don't see this as a good API design. Instead of passing the request options through the tus.Upload class, one could implement a child class of NodeHttpStack which sets the custom request directly. This custom NodeHttpStack can then be used by the `httpStack` option.

Alternatively, the default NodeHttpStack could have an additional method to set Node.js specific request options which are used for every request.

What do you think?

dragoscirjan:
I just remembered why I chose to pass the settings through `tus.Upload` and not through `HttpStack`. None of the `HttpStack` classes are visible outside of the package, and it would be redundant to copy the code outside of the project, just to extend.

I get what you were saying, though, so here is a new approach. This approach will export two `HttpStack` classes though. One which will allow passing options to `http(s).request` and another one which will force acceptance of insecure certificates no matter what.

I don't know whether I should update anything in the TypeScript definitions as well or not.

.travis.yml
```diff
@@ -3,8 +3,8 @@ dist: xenial
 language: node_js
 jobs:
   include:
-  - env: TEST_TARGET=browser
-    node_js: 12
+  # - env: TEST_TARGET=browser
```
Acconut, line 6: Is there a reason for disabling the browser tests?

lib/node/httpStack.js
```diff
@@ -5,22 +5,44 @@ import {Readable, Transform} from "stream";
...
+  }
+
   createRequest(method, url) {
-    return new Request(method, url);
+    return new Request(method, url, this._requestOptions);
   }
 
   getName() {
     return "NodeHttpStack";
+    // return this.constructor.name;
```
Acconut, line 18: Please remove this unused line.

lib/node/httpStack.js
```diff
@@ -5,22 +5,44 @@ import {Readable, Transform} from "stream";
...
+  }
+}
+
+/**
+ * A NodeHttpStack which will allow developer to customer `http(s).request()`
+ * options, yet force it to accept insecure certificates as well.
+ */
+export class InsecureNodeHttpStack extends NodeHttpStack {
```
Acconut, line 26: Frankly, I don't think we need to export an InsecureNodeHttpStack if the users can simply pass their own options to NodeHttpStack. That just increases the API surface with little benefit IMO.

lib/node/httpStack.js
```diff
@@ -47,10 +69,17 @@ class Request {
...
+        ...this._headers
+      };
       if (body && body.size) options.headers["Content-Length"] = body.size;
+      options = {
+        ...this._requestOptions,
+        ...options
+      };
 
-      let req = this._request = (options.protocol !== "https:") ? http.request(options) : https.request(options);
+      let req = (this._request = options.protocol !== "https:" ? http.request(options) : https.request(options));
```
Acconut, line 82: Can you explain why the change of these parenthesis is about?

lib/node/index.js
```diff
@@ -42,5 +42,7 @@ export {
   // Make FileUrlStorage module available as it will not be set by default.
   FileUrlStorage,
   canStoreURLs,
-  enableDebugLog
+  enableDebugLog,
+  HttpStack,
```
Acconut, line 46: :+1: Can you also add an export for the browser's HttpStack?

lib/upload.js
```diff
@@ -1,9 +1,11 @@
 /* global window */
-import DetailedError from "./error";
-import uuid from "./uuid";
 import { Base64 } from "js-base64";
+
```
Acconut, line 3: Can you please remove the empty lines?

test/spec/test-node-specific.js
```diff
@@ -39,7 +39,7 @@ describe("tus", function () {
       expect(err.message).toBe("tus: cannot automatically derive upload's size from input and must be specified manually using the `uploadSize` option");
     });
 
-    it("should reject streams without specifing the chunkSize", async function () {
+    it("should reject streams without specifying the chunkSize", async function () {
```
Acconut, line 42: Thank you very much :sweat_smile:

test/spec/test-node-specific.js
```diff
@@ -186,6 +186,13 @@ describe("tus", function () {
       await assertUrlStorage(storage);
     });
   });
+
+  describe("#InsecureNodeHttpStack", function () {
```
Acconut, line 190: Speaking of tests, do you know if it's possible to a test whether the options are properly passed to the request object?

.travis.yml
dragoscirjan: yes. I don't have an account & my tests would fail. will enable them back :)

lib/node/httpStack.js
dragoscirjan: wish I could've used that approach, however, I'm not sure if it's a TS feature or an ES feature. We can leave a string for now.

lib/node/httpStack.js
dragoscirjan: OK, I can leave with this :)

lib/node/httpStack.js
dragoscirjan: I have no clue what that was :)) reverted back

lib/upload.js
dragoscirjan: fixed

test/spec/test-node-specific.js
dragoscirjan: unfortunately no. most of the properties added to nodejs's `Request` are stored in private varibales

lib/node/index.js
dragoscirjan: Don't merge just yet. I'll try add a browser's HttpStack also, but it may take a little more time.

dragoscirjan:
Most comments are addressed. Give me some time to fix the HttpStack for browser as well :)

dragoscirjan:
OK. I've updated everything, including browser's HttpStack, but unfortunately, my browser tests fail; they failed even before I started changing the code

dragoscirjan:
Ha! They fail on my Travis account. The PR passed :)

dragoscirjan:
One more question. From the same discussion regarding exposing browser.HttpStack, I would expose Request and Response classes as well. There may be people who need them when extending HttpStack. What do you say?

PR #212, opened 2020-07-20
Allow Promise return type for onBeforeRequest and onAfterResponse.

Mentthew:


Acconut:
Thanks for the PR! Can you add a test case and update the documentation as well?

Mentthew:
> Thanks for the PR! Can you add a test case and update the documentation as well?

Done. ;)

Acconut (approved):
Thank you very much!

Issue #209, opened 2020-07-15
Terminate method doesnt execute options method onBeforeRequest


walterdavalos:
**Describe the bug**
When I invoke the method upload.abort(true) i noticed a DELETE request that did not send the parameters that I specified in the options method onBeforeRequest.

**To Reproduce**
1. Set for example onBeforeRequest to add a custom header
2. Start upload 
3. Abort upload with parameter shoultTerminate set true: upload.abort(true)
4. Check if the request has the custom header

**Setup details**
- Runtime environment: Node.js
- Used tus-js-client version: 2.1.1

Acconut:
Yes, you are right, this is missing. The problem is that the request handlers are not invoked, in the terminating function:

https://github.com/tus/tus-js-client/blob/2cad02ad6650a1b0aa231c4c3aa8829a8b742499/lib/upload.js#L118

Would you be willing to open a PR for this?

bedgerotto:
I created a [PR](https://github.com/tus/tus-js-client/pull/216) changing the terminate method to call onBeforeRequest and onAfterResponse from options, as it happens in the private method [_sendRequest.](https://github.com/tus/tus-js-client/blob/ffe21e2842fd243297fd3aba194c8d920170277d/lib/upload.js#L845) Hope it can be helpfull

Acconut:
Fixed in #216.

PR #216, opened 2020-08-20
 #209 Add onBeforeRequest and onAfterResponse to terminate method

bedgerotto:
Call onBeforeRequest and onAfterResponse inside the **terminate** method.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -227,6 +227,18 @@ onBeforeRequest: function (req) {
 }
 ```
 
+You can also return a Promise if you need to perform some calculations before the request is sent:
+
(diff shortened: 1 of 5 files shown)


Acconut:
Thank you very much for this PR! One small nitpick: The onAfterResponse callback should be invoked after every requests, regardless of the status code. With your PR, it will only be called if the status code is 204, but in the rest of tus-js-client, the callback is invoked also for error responses: https://github.com/tus/tus-js-client/blob/master/lib/upload.js#L841

Acconut (approved):
Thank you very much!

PR #217, opened 2020-08-27
Update index.d.ts

kabaliserv:
inconsistency between typescript typing and function definition


Acconut (approved):
You're right! Thank you!

PR #223, opened 2020-10-18
stop mixing XHTML with HTML5

G2G2G2G:


Acconut:
Thank you for the PR but I like keeping closing tags since they make the code easier to read. Or is there any pressing reason to remove them?

G2G2G2G:
Uhh, define an XHTML doctype if you want to use XHTML tags, then.
You didn't close your meta the same way, you also wrote this which is different:
https://github.com/tus/tus-js-client/blob/master/test/SpecRunner.html

To answer your question, no, since a lot of your code is messy there really isn't a reason.

Acconut:
> Uhh, define an XHTML doctype if you want to use XHTML tags, then.

That's not a very kind comment to make about a project that we provide to you for free.

> You didn't close your meta the same way, you also wrote this which is different

Good point, I wasn't aware of this inconsistency. I'll take care of it.

PR #224, opened 2020-10-21
Safer checking for unsupported callbacks

### LINKED ISSUE #222: A safer way to check for unsupported callbacks
**Is your feature request related to a problem? Please describe.**

I am working with [`vue-file-agent`](https://github.com/safrazik/vue-file-agent) which supports using `tus-js-client` as a plugin. It seems that it should be able to support version 1.x and 2.x at the same time with ***almost*** the same code, but with a few minor changes. It works almost perfectly; the specific issue that prevents it from working is the way that errors are thrown for callbacks. For example, `vue-file-agent` currently calls `abort()` with a callback as the second argument. This works for `tus-js-client` 1.x, but fails for 2.x.

**Describe the solution you'd like**

I need a way to check whether the version of `tus-js-client` that is loaded is the old callback-based version or the new promise-based version.

It should be very simple to do this by changing two lines in each function that throws such an error:

For example, change this:

```
  abort(shouldTerminate, cb) {
    if (typeof cb === "function") {
      throw new Error("tus: the abort function does not accept a callback since v2 anymore; please use the returned Promise instead");
    }
```

to this:

```
  abort(shouldTerminate) {
    if (arguments.length > 1 && typeof arguments[1] === "function") {
      throw new Error("tus: the abort function does not accept a callback since v2 anymore; please use the returned Promise instead");
    }
```

This would allow me to do something simple like this:

```
if (upload.abort.length === 2) then {
    upload.abort(false, callback);
} else {
    upload.abort(false).then(callback);
}
```

**Describe alternatives you've considered**

`tus-js-client` does not provide any other way that I could find to detect the version.

The only workaround I can think of is to use a `try`/`catch` like this, but it doesn't make sense to need to do that:

```
try {
    upload.abort(false, callback);
} catch(err) {
    upload.abort(false).then(callback);
}
```

**Can you provide help with implementing this feature?**

I would be happy to submit a PR if you'd like.

kohenkatz:
fixes #222 

This allows implementers to support both tus-js-client 1.x and 2.x, by allowing them to check how many arguments the `abort()` and `terminate()` functions receive.

It will still throw an error if the callback is provided (as the second argument of `abort()` and as the second or third argument of `terminate()`).

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -109,9 +109,17 @@ class BaseUpload {
    * @param {object} options Optional options for influencing HTTP requests.
    * @return {Promise} The Promise will be resolved/rejected when the requests finish.
    */
-  static terminate(url, options = {}, cb) {
-    if (typeof options === "function" || typeof cb === "function") {
+  static terminate(url, options) {
+    // Count the number of arguments to see if a callback is being provided as the last
+    // argument in the old style required by tus-js-client 1.x, then throw an error if it is.
+    // `arguments` is a JavaScript built-in variable that contains all of the function's arguments.
+    if (arguments.length > 1 && typeof arguments[arguments.length - 1] === "function") {
       throw new Error("tus: the terminate function does not accept a callback since v2 anymore; please use the returned Promise instead");
+    }
+    // Note that in order for the trick above to work, a default value cannot be set for `options`,
+    // so the check below replaces the old default `{}`.
+    if (options === undefined) {
+      options = {};
     }
 
     const req = openRequest("DELETE", url, options);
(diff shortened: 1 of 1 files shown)


Acconut (approved):
Thank you very much!

Acconut:
I am going to merge this already but would you mind explaining what you mean with this comment:

> Note that in order for the trick above to work, a default value cannot be set for `options`, so the check below replaces the old default `{}`.

kohenkatz:
@Acconut The `Function.length` property only counts arguments until it finds an argument with a default value.

An example from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length) is exactly what we are doing:

```
console.log((function(a, b = 1, c) {}).length);
// 1, only parameters before the first one with 
// a default value is counted
```

If we modify that example to remove the default on `b`, then we get this:

```
console.log((function(a, b, c) {}).length);
// 3
```

If we then remove the third argument (as this PR did), we get this:

```
console.log((function(a, b) {}).length);
// 2
```

It is now possible to check whether we are using the Promise version or the callback version like this:

```
if (upload.terminate.length === 2) {
    // We are using the promise version
} else {
    // We are using the callback version
}
```

Acconut:
Thank you very much for the explanation, @kohenkatz!

PR #234, opened 2021-01-27
Fixed HttpStack type to class

kqito:
`HttpStack` definition was defined as `interface`, fixed to define `class` it.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1,5 +1,7 @@
 // Type definitions for tus-js-client
 
+import { RequestOptions } from "https";
+
 export const isSupported: boolean;
 export const canStoreURLs: boolean;
 export const defaultOptions: UploadOptions;
@@ -83,7 +85,8 @@ interface SliceResult {
   done: boolean;
 }
 
-export interface HttpStack {
+export class HttpStack {
+  constructor(requestOptions?: RequestOptions);
   createRequest(method: string, url: string): HttpRequest;
   getName(): string;
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1590,6 +1590,12 @@
       "integrity": "sha1-aaI6OtKcrwCX8G7aWbNh7i8GOfY=",
       "dev": true
     },
+    "@types/node": {
+      "version": "14.14.22",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-14.14.22.tgz",
+      "integrity": "sha512-g+f/qj/cNcqKkc3tFqlXOYjrmZA+jNBiDzbP3kH+B+otKFqAdPgVTGP1IeKRdMml/aE69as5S4FqtxAbl+LaMw==",
+      "dev": true
+    },
     "@types/normalize-package-data": {
       "version": "2.4.0",
       "resolved": "https://registry.npmjs.org/@types/normalize-package-data/-/normalize-package-data-2.4.0.tgz",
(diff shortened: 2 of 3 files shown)


Acconut:
Why was this change necessary? This type is intended to by an interface so people can pass their own HTTP stacks to the `httpStack` option: https://github.com/tus/tus-js-client/pull/234/files#diff-cfc7807f6b174e1c3705c5c2747f56e49a889a52999dc96e4080a3af0131a534R53

kqito:
Thanks for the reply!

I see.

For example, when I try to upload from `http://localhost` to `https://sample.com` using this library, the destination protocol of the PATCH request is `http`.(I have to deal with this for CORS)

I am not sure if this is a tus specification or a bug in this library.

In order to convert the `http` to `https`, I wanted to extend the existing HttpStack, so I created this PR.

If there's another way, I'd appreciate it if you could let me know!

Acconut:
The issue you are seen is not a bug in tus-js-client but instead very often the result of an incomplete configuration of your tus server or any proxy. I do not know what tus server you are using, but tusd has a description for how to set it up properly: https://github.com/tus/tusd/blob/master/docs/faq.md#can-i-run-tusd-behind-a-reverse-proxy

I recommend you to fix your configuration instead of only patching it in your tus-js-client code :)

kqito:
Thanks for the reply !

I'll try that, thanks.

I'd like to discuss one last thing with you.

https://github.com/tus/tus-js-client/blob/master/lib/node/index.js#L46

When I import this HttpStack `class` in a TypeScript environment, it is recognized as an HttpStack interface, and I can't use the `class`.

e.g.
```
import { HttpStack } from 'tus-js-client'

const httpStack = new HttpStack()
```

error message
```
error TS2693: 'HttpStack' only refers to a type, but is being used as a value here.
```

What should I do if I want to use it as a class?

Acconut:
Ah yes, I see what you are talking about! Thanks for bringing it up! We define `HttpStack` as an interface in `index.d.ts` (https://github.com/tus/tus-js-client/blob/master/lib/index.d.ts#L86) but this then overwrites the `HttpStack` class we export: https://github.com/tus/tus-js-client/blob/master/lib/node/index.js#L46

A solution would be to change either the name of the interface or of the exported class, so this conflict does not occur anymore. Would you mind opening a PR for this?

kqito:
I'm sorry for my bad English :)

You're right ! 

> A solution would be to change either the name of the interface or of the exported class, so this conflict does not occur anymore. Would you mind opening a PR for this?

Ok, I'll create PR for that👍
Please give me some time.

kqito:
Hello!

> A solution would be to change either the name of the interface or of the exported class, so this conflict does not occur anymore. Would you mind opening a PR for this?

I'm working on this PR thing, and I need to talk to you about one thing!

In front of this you are

> This type is intended to by an interface so people can pass their own HTTP stacks to the httpStack option


You said so, but I thought that if this PR (#234) is merged, `HttpStack` will be available as both class and interface.

For example
```ts
// export class HttpStack {
//   constructor(requestOptions?: RequestOptions);
//   createRequest(method: string, url: string): HttpRequest;
//   getName(): string;
// }


// as interface
const httpStack: HttpStack = {
  createRequest: (method: string, url: string) => 'httpRequest' as any,
  getName: () => ''
}


// as class
class CustomHttpStack extends HttpStack {}

```

Issue #239, opened 2021-02-03
401 on HEAD call resets file progress


dandenton:
**Describe the bug**
Due to weirdness with our reverse proxy we occasionally have http/2 or CORS exceptions returned while uploading a chunk (PATCH) instead of a 401.  On that failure it's retrying the HEAD request, receiving a 401 which then resets the file.  After re-authentication the file starts over rather than resuming.

Screenshot shows "Finished" just because of the HAR file but the original browser showed a CORS exception.
![image](https://user-images.githubusercontent.com/56007134/106806230-eadadd80-6624-11eb-8ccd-88fb2e914e3c.png)

**To Reproduce**
1. Start a file upload and have more than 1 chunk upload
2. Encounter an exception on a chunk that results in a retry (network exception or disconnect during the next PATCH)
3. Receive a 401

**Expected behavior**
It looks like any 4XX response for the HEAD request removes the file from storage.  Can we check for 401/403 errors and exclude that from the logic at https://github.com/tus/tus-js-client/blob/master/lib/upload.js#L605

![image](https://user-images.githubusercontent.com/56007134/106812173-87ed4480-662c-11eb-9b8e-ac8906939011.png)

**Setup details**
- Reproduced in Chrome and Edge
- Used tus-js-client version: 2.2.0
- Server is using tusdotnet with authentication

Acconut:
> Instead, it would be good if TUS throws an authentication error and lets the user decide how they want to handle it.

Handling a 401 different than other 4XX status codes is not ideal in my opinion. I would prefer a more generic option, such as `removeFingerprintOn4xx`, which controls whether a 4xx response causes the fingerprint to be removed, similar to how we have `removeFingerprintOnSuccess`. You could then disable this setting, catch the error using `onError` and decide how to proceed. Does this make sense?

amolghode1981:
@Acconut Thanks for the suggestion. That would help. 

For the argument's sake, I would still prefer having that control in the user's hand specifically for the authentication failure. TUS is very powerful for chunk uploads and resumeability. There are scenarios where user would work on the local copy and upload it in parallel. And also that they may want to do it across multiple browser sessions. So auth expiry is a common phenomenon. Letting the user authenticate and resume (instead of restarting the upload from the first byte) would reduce the pain by a bit IMHO.

Nevertheless, you are in a better situation to make a judgment on this. Thanks once again for agreeing to add the `removeFingerprintOn4xx` flag. :)

Acconut:
[...]

PR #240, opened 2021-02-03
Update _resumeUpload to resume unauthenticated requests

dandenton:
Proposed fix for https://github.com/tus/tus-js-client/issues/239


Acconut:
Thank you very much for the contribution. However, I think this is a very application-specific change and without being configurable not very useful for most users of tus-js-client.

Since I responded so late to this PR (my apologies), I am going to close this PR. However, if you are interested in bringing this feature in a more configurable way into tus-js-client, let us know :)

PR #241, opened 2021-02-06
Rename HttpStack to DefaultHttpStack

kqito:
## Overview
There was a bug that the exported [HttpStack class](https://github.com/tus/tus-js-client/blob/master/lib/node/index.js#L46) conflicts with the [HttpStack interface](https://github.com/tus/tus-js-client/blob/master/lib/index.d.ts#L86), preventing it from importing properly in a typescript environment.

so, I fixed this bug by renaming `HttpStack` to `DefaultHttpStack`

https://github.com/tus/tus-js-client/pull/234#issuecomment-771880925

## Question
Should I also rename [this HttpStack](https://github.com/tus/tus-js-client/blob/master/lib/browser/index.js#L45) in `lib/browser/index.js`?


Acconut:
Thank you, merged in https://github.com/tus/tus-js-client/commit/bafe931887f4e46dfd2bb31d653761cc1cb4b912 with some additional changes.

PR #243, opened 2021-02-16
Improve stream type detection

### LINKED ISSUE #229: stream needs to be instanceof Readable
**Describe the bug**

Error `source object may only be an instance of Buffer or Readable in this environment` when trying to upload a stream which is not directly an instaceof `Readable`.

**To Reproduce**
```bash
yarn add tus-js-client@1.8.0 into-stream
```

index.js
```javascript
const tus = require('tus-js-client')
const intoStream = require('into-stream')

// into-stream returns a stream where instanceof stream !== Readable
const stream = intoStream('my test string')

const tusUpload = new tus.Upload(stream, {
  endpoint: 'http://localhost:1234',
  resume  : true,
  metadata: {
    fieldname: 'field',
  },
  onError: console.error,
  onSuccess: () => console.log('done'),
})

tusUpload.start()
```

```bash
node index.js

Error: source object may only be an instance of Buffer or Readable in this environment
    at getSource (node_modules/tus-js-client/lib.es5/node/source.js:234:12)
```

**Expected behavior**
Should upload the stream. I'm not sure if it's a bug but I think so, because other libraries handles these kind of streams without problems. For instance [form-data](https://github.com/form-data/form-data/) uploads it fine.

I think `instanceof Readable` is not the best way to determine if an object is a readable stream. See:
https://stackoverflow.com/questions/23885095/nodejs-check-if-variable-is-readable-stream

Maybe we can instead do something like this? https://github.com/sindresorhus/is-stream/blob/master/index.js

https://github.com/transloadit/node-sdk/issues/86

**workaround hack**

```js
const rawStream = intoStream(str)
const stream = new require('stream').PassThrough();
rawStream.pipe(stream);
```

**Setup details**
Please provide following details, if applicable to your situation:
- Runtime environment: Node.js
- Used tus-js-client version: 1.8.0

mifi:
by using `into-stream`
Fixes #229

### DIFF
--- a/lib/node/fileReader.js
+++ b/lib/node/fileReader.js
@@ -1,5 +1,6 @@
-import {Readable, Transform} from "stream";
+import {Transform} from "stream";
 import {ReadStream, createReadStream} from "fs";
+import isStream from "is-stream";
 
 class BufferSource {
   constructor(buffer) {
@@ -170,7 +171,7 @@ export default class FileReader {
       return Promise.resolve(new FileSource(input));
     }
 
-    if (input instanceof Readable) {
+    if (isStream.readable(input)) {
       return Promise.resolve(new StreamSource(input, chunkSize));
     }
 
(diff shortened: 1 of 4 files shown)


Acconut (approved):
Thank you very much!

PR #249, opened 2021-03-15
Update Linter

kvz:
Discussion in https://github.com/transloadit/eslint-config-transloadit/issues/1

Some background, we used to have standardjs linting (still do) but @mifi noted that airbnb has more safetynet features. I have to concur, it can prevent many more bugs. But changing would also brings a lot of stylistic changes that we're probably not welcoming in many of our js code bases.

So that gave birth to the idea of the transloadit eslint config, which bases off of airbnb, then makes stylistic modifications to pretty much suit what we had with standardjs. Giving us more safetly while having roughly the same code as before.

For the tus-js-client repo, we did not use standardjs before, so the changes will be more drastic.

Not all the safety features can be automatically fixed by eslint. If we had to do it in a single PR, it would grow so large and be open for so long, that it just invites merge conflicts, and we might never merge. Instead, the diff between old and new is listed in `.eslintrc.json` as warnings. So they would already appear yellow in your editor, but they do not block CI. We can then, for the rules we agree with, remove them from the eslintrc and make the appropriate code changes, one by one, in smaller PRs. 

For any pending PRs, to reduce merge conflicts, we should follow something along the lines of: https://gist.github.com/kvz/b205470849f0489549f43c7a27a406dd

This PR also switches over to yarn for speed advantages. In all openness i found npm to be unbearable in setting this up.

## update

Waiting a long time for Travis, if it will even start working. Perhaps we should first open a PR for switching to GHA, then afterwards merge this one.


kvz:
With Ife gone and Marius on very few hours, likely this won't get a review. Team agreed to the changes in https://github.com/transloadit/eslint-config-transloadit/issues/1 and all other depending projects however so i'll go ahead and merge this once I get CI to pass. Took care of moving to GHA first in https://github.com/tus/tus-js-client/pull/250 so CI is a possibility again

PR #251, opened 2021-03-25
Add call to findPreviousUploads to readme

smatsson:
Seems like the call to `findPreviousUploads` to continue any existing upload is missing from the example so I've added it based on https://github.com/tus/tus-js-client/blob/master/docs/usage.md#example-let-user-select-upload-to-resume

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,16 @@ input.addEventListener("change", function(e) {
         }
     })
 
-    // Start the upload
-    upload.start()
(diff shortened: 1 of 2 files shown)


Acconut (approved):
Thank you very much! I just reverted the Vimeo example because `findPreviousUploads` is irrelevant if `uploadUrl` is set.

smatsson:
Thank you. TIL :)

PR #265, opened 2021-06-22
Cast options.chunkSize to integer

oliverpool:
Hi,

I just had a bug, where I set the `chunkSize` to `"100"`.

Everything was fine on the first chunk, but after that the chunk had a size of `100100` bytes :laughing: 

I have fixed my app to cast the config to an int, but I think that adding this failsafe upstream shouldn't do any harm (and it does not affect `Infinity === +Infinity`)

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -714,7 +714,7 @@ class BaseUpload {
    */
   _addChunkToRequest (req) {
     const start = this._offset
-    let end = this._offset + this.options.chunkSize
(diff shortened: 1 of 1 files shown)


Acconut:
Thank you for this PR and apologies for my late reply! A failsafe sounds reasonable! However, it would be more advantageous to place the casting in the constructor (https://github.com/tus/tus-js-client/blob/master/lib/upload.js#L50), so all other code can benefit from safety check as well. What do you think?

PR #297, opened 2021-10-29
enable all eslint rules and fix problems

mifi:
See discussion in #291
- had to disable no-use-before-define many places where onSuccess uses the upload object, maybe in the future onsuccess could provide upload as an argument to onSuccess
- also a lot of functions moved (as eslint doesn't allow function hoisting)
- disabled eslint for demos/cordova and demos/browser
- disabled import/no-extraneous-dependencies for test code (which uses dev dependencies)

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -273,8 +273,7 @@ class BaseUpload {
 
     // Count the number of arguments to see if a callback is being provided as the last
     // argument in the old style required by tus-js-client 1.x, then throw an error if it is.
-    // `arguments` is a JavaScript built-in variable that contains all of the function's arguments.
-    if (arguments.length > 1 && typeof args[arguments.length - 1] === 'function') {
+    if (args.length > 1 && typeof args[args.length - 1] === 'function') {
       throw new Error('tus: the terminate function does not accept a callback since v2 anymore; please use the returned Promise instead')
     }
     // Note that in order for the trick above to work, a default value cannot be set for `options`,
@@ -562,8 +561,7 @@ class BaseUpload {
     const [shouldTerminate] = args
     // Count the number of arguments to see if a callback is being provided in the old style required by tus-js-client 1.x,
     // then throw an error if it is.
-    // `arguments` is a JavaScript built-in variable that contains all of the function's arguments.
-    if (arguments.length > 1 && typeof args[1] === 'function') {
+    if (args.length > 1 && typeof args[1] === 'function') {
       throw new Error('tus: the abort function does not accept a callback since v2 anymore; please use the returned Promise instead')
     }
 


mifi:
any chance of getting this merged @Acconut before we get too many conflicts? Or are there any rules you don't agree with that you would like to disable?

bin/browserstack-jasmine.js
```diff
@@ -1,5 +1,6 @@
 /* eslint no-console: 0 */
 
+// eslint-disable-next-line import/no-extraneous-dependencies
```
Acconut, line 3: Do you know why import/no-extraneous-dependencies has to be disabled here? browserstack-runner is defined as a dependency in package.json

bin/puppeteer-jasmine.js
```diff
@@ -6,6 +6,7 @@
  * according to the result.
  */
 
+// eslint-disable-next-line import/no-extraneous-dependencies
```
Acconut, line 9: Again, is it necessary to disable import/no-extraneous-dependencies here?

lib/node/urlStorage.js
```diff
@@ -140,15 +159,15 @@ export class FileUrlStorage {
     readFile(this.path, 'utf8', (err, data) => {
       if (err) {
         // return empty data if file does not exist
-        err.code === 'ENOENT' ? cb(null, {}) : cb(err)
+        if (err.code === 'ENOENT') cb(null, {})
+        else cb(err)
       } else {
         try {
-          data = !data.trim().length ? {} : JSON.parse(data)
+          const data2 = !data.trim().length ? {} : JSON.parse(data)
```
Acconut, line 166: I assume there is a rule disallowing the reassignment of arguments. Could we loosen up this rule? I actually think the previous code where reassignment is allowed is more reasonable. Or is there a specific reason against it?

lib/upload.js
```diff
@@ -112,15 +268,18 @@ class BaseUpload {
...
    */
-  static terminate (url, options) {
+  static terminate (...args) {
+    const [url, optionsIn] = args
+
     // Count the number of arguments to see if a callback is being provided as the last
     // argument in the old style required by tus-js-client 1.x, then throw an error if it is.
     // `arguments` is a JavaScript built-in variable that contains all of the function's arguments.
-    if (arguments.length > 1 && typeof arguments[arguments.length - 1] === 'function') {
+    if (arguments.length > 1 && typeof args[arguments.length - 1] === 'function') {
```
Acconut, line 277: Could you explain this change? Are we not allowed to access `arguments`, but we can use `arguments.length`?

Acconut (commented):
Thank you for reminding me of this PR! I only have a few questions about it

lib/upload.js
mifi: TBH I can't explain that 😅 It seems like [the rule](https://eslint.org/docs/rules/prefer-rest-params) doesn't enforce all uses of arguments, I don't know why. I'll remove the other uses too. Or if you think this rule should be disabled we can do that too.

Acconut:
Implement in https://github.com/tus/tus-js-client/pull/390 which fixed the merge conflicts. Thanks for your work, @mifi!

Issue #298, opened 2021-11-05
Custom `splitSizeIntoParts` function


m-cat:
# Feature Request

**Is your feature request related to a problem? Please describe.**
In our project we needed to customize the size of the chunks when `parallelUploads` is used. This is because our backend expects chunks of a certain size.

**Describe the solution you'd like**
I forked the repo so that I could change the `splitSizeIntoParts` function. However, I'd like to make this function customizable and merge that change upstream, so that our project can keep depending on the main repo for tus client.

**Describe alternatives you've considered**
None

**Can you provide help with implementing this feature?**
Yes, I want to write code allowing passing a custom `splitSizeIntoParts` and raise a PR with the change.

**Additional context**
None

PR #299, opened 2021-11-05
Add `parallelUploadBoundaries` option and test

m-cat:
# Pull Request

## Overview

Adds `parallelUploadBoundaries` option to customize the size of parallel upload chunks.

## Use Case

In our project we needed to customize the size of the chunks when parallelUploads is used. This is because our backend expects chunks of a certain size.

## Test

Not sure how the devs here test PRs. Here is a simple application that I built on top of the changes:

<snip, will rebuild on request>

## Closes

Closes https://github.com/tus/tus-js-client/issues/298

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -29,6 +29,7 @@ const defaultOptions = {
   chunkSize                  : Infinity,
   retryDelays                : [0, 1000, 3000, 5000],
   parallelUploads            : 1,
+  splitSizeIntoParts         : null,
   storeFingerprintForResuming: true,
   removeFingerprintOnSuccess : false,
   uploadLengthDeferred       : false,
@@ -96,6 +97,10 @@ class BaseUpload {
     // An array of BaseUpload instances which are used for uploading the different
     // parts, if the parallelUploads option is used.
     this._parallelUploads = null
+
+    // A custom function for splitting the upload size into parts, if the
+    // parallelUploads option is used.
+    this._splitSizeIntoParts = null
 
     // An array of upload URLs which are used for uploading the different
     // parts, if the parallelUploads option is used.
@@ -196,6 +201,12 @@ class BaseUpload {
       })
     }
 
+    if (this.options.splitSizeIntoParts) {
+      if (this.options.parallelUploads <= 1) {
+          this._emitError(new Error(`tus: cannot use the splitSizeIntoParts option when parallelUploads is disabled`))
+      }
+    }
+
     this.options.fingerprint(file, this.options)
       .then((fingerprint) => {
         if (fingerprint == null) {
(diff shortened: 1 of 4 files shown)


Acconut:
Thanks for the PR, but I think the callback function is too much complexity. What about just passing the array of the part bounds directly in `parallelUploads`? For example,

```js
parallelUploads: [{start: 0, end: 99}, ..]
```

That would make the code a lot more simpler since no additional option is required.

In addition, please also add any changed parameter to the documentation in the docs folder.

m-cat:
Hi @Acconut, thanks for the reply. I'm not sure whether passing the final parts or passing a function is better. I'm personally leaning towards keeping the function as it's one less step for the user, and so less prone to error, but I'll trust your judgment on that one.

I don't agree with overloading `parallelUploads`, though. That doesn't seem less complex than making a new `parts` or `parallelParts` option - it's just stuffing the complexity into an existing option. Having a new option seems like it would be more clear, to me at least.

Acconut:
> I'm personally leaning towards keeping the function as it's one less step for the user, and so less prone to error, but I'll trust your judgment on that one.

I tend to disagree :) Callbacks are more error-prone in my experience as the order of execution is not as clear as if the user has to do the computation up-front. In addition, since the stack trace for any exception in the callback will also contain lines from tus-js-client, users easily mistaken that for bugs in tus-js-client, inside of inside their own code.

Furthermore, tus-js-client currently only uses callback for options when the property cannot be pre-computed before starting the upload (i.e. onShouldRetry). All other options are pre-computed and passed directly as values. I think we should be consistent and keep this pattern.

> Having a new option seems like it would be more clear, to me at least.

True, I agree. A new option `parallelUploadBoundaries` would be better. However, we should add validation to ensure that `parallelUploads === parallelUploadBoundaries.length`.

What do you think? Are you interested in making these changes to your PR?

m-cat:
That makes sense to me, I can make those changes. I've been on vacation, but I can try to get to this in the next week or so.

Acconut:
Great, let me know if you need assistance.

lib/upload.js
```diff
@@ -974,11 +995,10 @@ function resolveUrl (origin, link) {
  *
  * @param {number} totalSize The byte size of the upload, which will be split.
  * @param {number} partCount The number in how many parts the upload will be split.
- * @param {string[]} previousUrls The upload URLs for previous parts.
  * @return {object[]}
  * @api private
  */
-function splitSizeIntoParts (totalSize, partCount, previousUrls) {
+function splitSizeIntoParts (totalSize, partCount) {
```
m-cat, line 1001: Didn't see a need for `previousUrls` to be a parameter here, so I factored it out for better separation of concerns. I can revert this change since my PR is not using a callback anymore based off this function. I think the change is an improvement, though.

m-cat:
Done @Acconut. Feedback welcome!

test/spec/test-parallel-uploads.js
```diff
@@ -178,6 +197,190 @@ describe('tus', () => {
       expect(testUrlStorage.removeUpload).toHaveBeenCalled()
     })
 
+    it('should split a file into multiple parts based on custom `parallelUploadBoundaries`', async () => {
```
Acconut, line 200: I see that you copied the test case from above but I think we can remove some details form this test as they are already covered in the test case above. I am mainly thinking about removing testUrlStorage, retryDelays (plus the 500 response on line 329), the headers option, the metadata option, the onProgress callback. Could you do this?

m-cat: Done!

Acconut:
Merged in https://github.com/tus/tus-js-client/commit/169b099eae1cab3bf9b98e8dcdcf7ffca14c5e6b with some additional changes. Thanks for your work.

Issue #238, opened 2021-02-02
findPreviousUploads returns empty array despite the file and the uploadUrl being the same.


JulianH99:
**Describe the bug**
I'm uploading a video with tus to vimeo and I have been asked to implement the `findPreviousUploads` functionality. But, for example, If I'm uploading a file called  `video.mp4` and close the tab, and then come back to upload the same file again, the array of previous uploads is empty. I know vimeo provides it's own uploadURL for every upload, but in this case I'm saving the uploadURL of the file in localStorage so that I can retrieve it later when the user uploads the same file again.

**Expected behavior**
The array of previous uploads should not be empty since I'm uploading the same file to the same uploadURL

**Setup details**
Please provide following details, if applicable to your situation:
- Runtime environment: Browser
- Used tus-js-client version: 2.2.0
- Used tus server software: vimeo

Acconut:
This is not a bug but the intended behavior instead. tus-js-client will only store upload URL in localStorage if it created the upload. However, since in your case Vimeo already created the upload, and you are only passing in the `uploadUrl` option, tus-js-client will not store any URL. Since it does not store anything, `findPreviousUploads` is not able to retrieve any upload.

FYI, here is the code for storing the upload URL: https://github.com/tus/tus-js-client/blob/e145a0521f92819bb5666b7fa17532267cf1b1b3/lib/upload.js#L578

Here is the documentation stating that only created uploads will be saved: https://github.com/tus/tus-js-client/blob/master/docs/api.md#storefingerprintforresuming

As a workaround I see two options:
- Save the upload URL from Vimeo on your own in localStorage, retrieve it on your own, and then pass it to tus-js-client again when resuming.
- Retrieve the default UrlStorage and call the `addUpload` method to add the upload URL from Vimeo on your own: https://github.com/tus/tus-js-client/blob/master/docs/api.md#urlstorage

JulianH99:
Thanks for answering. I'm currently trying the first option but still no luck, upload starts from 0 again and does not seem to be resuming anything. Any clues on what could be wrong? this is my code:
```typescript
 const uploadObject = new Upload(file, {
      uploadUrl: this.uploadUrl,
      overridePatchMethod: false,
      retryDelays: [0, 3000, 5000, 10000, 20000],
      metadata: {
        filename: file.name,
        filetype: file.type,
      },

);
```

`this.uploadUrl` is the previously returned url from the localStorage

Acconut:
[...]

PR #323, opened 2022-03-05
Adds function to store uploadUrl

### LINKED ISSUE #200: Resume vimeo video not working
**Question**
Hi all,

I am trying to enable resuming a video on page refresh/connection lost. Right now, when I refresh a page, it restarts from the beginning instead of resuming.

Here is the code snippet:
```typescript
 this.upload = new Upload(this.videoFile, {
      endpoint: undefined,
      uploadUrl: this.model.video.uploadUrl,
      resume: true,
      retryDelays: [0, threeSeconds, fiveSeconds, tenSeconds, twentySeconds],
      onError: (error: Error): void => {
        this.onError(error.message);
      },
      onProgress: (bytesUploaded: number, bytesTotal: number): void => {
        this.progress = Math.round((bytesUploaded / bytesTotal) * percent);
      },
      onSuccess: (): void => {
        console.log('success');
      }
    });
```

- The videoFile here is saved in localStorage so that on refresh, we can retrieve it.
- The uploadUrl is always the same. It is saved in the database.
- When I do not provide an uploadUrl, I get a 405 error saying that POST is not an allowed method for https://us-files.tus.vimeo.com/files/...

Please let me know what I am not seeing. Thank you again in advance!

**Setup details**
Please provide following details, if applicable to your situation:
- Runtime environment: Browser, Angular 9
- Used tus-js-client version: 1.8.0

### LINKED ISSUE #301: Resumable Vimeo upload not working with tus js?
Hi all,

I have a pretty simple setup - user picks a video from the gallery, we convert the video into a blob, then we create the Vimeo upload url (which i have confirmed to be consistent), then we create an `Upload()` object to upload the video via tus to Vimeo:

```js
   const blob = await fetch(`data:video/mp4;base64,${b64}`).then((res) =>
      res.blob()
    );
    const uploadLink = await this.generateVideoUploadUrl(blob.size);
    let fileName = this.path.substring(this.path.lastIndexOf('/') + 1);

    const upload = new Upload(blob, {
      endpoint: uploadLink,
      uploadUrl: uploadLink,
      retryDelays: [0, 1000, 3000, 5000],

      onError: function (error) {
        console.log('Failed because: ' + error);
      },
      onProgress: function (bytesUploaded, bytesTotal) {
        console.log('progress');
        var percentage = ((bytesUploaded / bytesTotal) * 100).toFixed(2);
        console.log(bytesUploaded, bytesTotal, percentage + '%');
      },
      onSuccess: function () {
        console.log('Download %s from %s', upload.file, upload.url);
      },
    });

    upload.start();
```

However, I get nothing in terms of the video being uploaded or any signs in the console. Any reason as to why this could be?

Please provide following details, if applicable to your situation:
- Runtime environment: Ionic 5 with Capacitor 
- Used tus-js-client version: tus-js-client@2.3.0
- Used tus server software: using Vimeo

stebogit:
Adds the `storeUploadUrlForResume` public method to store the `uploadUrl`, useful when creating the upload URL from a separate server like Vimeo.
It also adds some details regarding using Vimeo API.

Fixes #200
Fixes #301
Ref #238

### DIFF
--- a/demos/browser/index.html
+++ b/demos/browser/index.html
@@ -16,7 +16,7 @@
 
       <p>
         For a prettier demo please go to the
-        <a href="http://tus.io/demo.html">tus.io</a> website.
+        <a href="https://tus.io/demo.html">tus.io</a> website.
         This demo is just here to aid developers.
       </p>
 
@@ -73,7 +73,7 @@
       <div class="row">
         <div class="span8">
           <div class="progress progress-striped progress-success">
-            <div class="bar" style="width: 0%;"></div>
+            <div class="bar" style="width: 0;"></div>
           </div>
         </div>
         <div class="span4">
@@ -84,7 +84,7 @@
       <hr>
       <h3>Uploads</h3>
       <p id="upload-list">
-        Succesful uploads will be listed here. Try one!
+        Successful uploads will be listed here. Try one!
       </p>
 
     </div>
--- a/docs/api.md
+++ b/docs/api.md
@@ -424,7 +424,8 @@ tus.Upload.terminate(url).then(function () {
 
 ## tus.Upload#storeUploadUrlForResume(uploadUrl)
 
-Stores the provided `uploadUrl` to the URL storage using the input file's fingerprint to allow later resume (see `tus.Upload#findPreviousUploads` or `tus.Upload#resumeFromPreviousUpload`). This method can be useful when your tus server automatically creates the upload resources, like when interacting with the Vimeo API.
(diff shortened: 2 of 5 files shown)


Acconut:
Thank you for this PR! To be frank, I am not sure if this is the best approach. If I understand correctly, the reason for this PR is that tus-js-client does not store the upload URL if `uploadUrl` is specified. However, what if we simply change that and store the `uploadUrl` in these cases as well. Then this function from your PR is not necessary, which would avoid complexity and confusion for users.

I am not a 100% sure if this would be correct approach. What do you think?

Acconut:
This is implemented in https://github.com/tus/tus-js-client/issues/371.

PR #327, opened 2022-03-21
Export DetailedError from package

vkrol:
For example, it's useful for checking in the `shouldRetry` callback:
```js
function onShouldRetry(error) {
  if (error instanceof DetailedError) {
    return doSomethingWithOriginalResponse(error.originalResponse);
  }
}
```

Also this class is already exported, but only in types :) https://github.com/tus/tus-js-client/blob/dc564f143d3eab2b5ddd0848822662ad80606058/lib/index.d.ts#L115


Acconut (approved):
Thank you, you are right!

PR #347, opened 2022-04-27
Work In Progress - feat: checksum extension

manohar27:
Adds Checksum Extension support. 

This is work in progress.
Uses SubtleCrypto on browser and crypto library on nodejs


Acconut:
Thank you very much for this PR! It looks interesting. Let me know once it is ready for a review!

manohar27:
> Thank you very much for this PR! It looks interesting. Let me know once it is ready for a review!

Sorry this took so long. 

Need some help with code reviews and testing this

Acconut:
Thank you! I will try to review this in the next two weeks!

lib/browser/extensions/checksum.js
```diff
@@ -0,0 +1,39 @@
...
+  constructor (algo = 'SHA-256') {
+    this.algo = this._resolveAlgo(algo)
+  }
+
+  /**
+   * Resolves algorithm to one of the inbuilt types
+   * @param {String} algo contains user provided checksumAlgo option
+   */
+  _resolveAlgo = (algo) => {
+    const resolvedAlgo = Checksum.supportedAlgorithms.find(supportedAlgo => supportedAlgo === algo.toUpperCase())
```
Acconut, line 13: That logic could be simplified using `Array#includes`. All in all, the _resolveAlgo method seems a bit unnecessary, as this could all be done in the constructor directly.

lib/browser/extensions/checksum.js
```diff
@@ -0,0 +1,39 @@
...
+    return resolvedAlgo
+  }
+  /**
+   * Gets Hexadecimal digest using the algorithm set in this.algo
+   * @param {ArrayBuffer} data contains the chunk of data to be hashed
+   */
+
+  getHexDigest = async (data) => {
+    try {
+      const hashBuffer = await crypto.subtle.digest(this.algo, data)
```
Acconut, line 28: A method for checking if the Crypto API is support would be great, so the users gets notified early if that's not the case.

lib/browser/extensions/checksum.js
```diff
@@ -0,0 +1,39 @@
...
+      const hashArray = Array.from(new Uint8Array(hashBuffer)) // convert buffer to byte array
+      const hashHex = hashArray
+        .map((b) => b.toString(16).padStart(2, '0'))
+        .join('') // convert bytes to hex string
+      return hashHex
+    } catch (err) {
+      throw new Error('tus: could not compute checksum for integrity check')
```
Acconut, line 35: Let's forward the error message here.

lib/node/extensions/checksum.js
```diff
@@ -0,0 +1,38 @@
...
+  constructor (algo = 'sha256') {
+    this.algo = this.resolveAlgo(algo)
+  }
+
+  /**
+   * Resolves algorithm to one of the inbuilt types
+   * @param {String} algo contains user provided checksumAlgo option
+   */
+  resolveAlgo = (algo) => {
+    const resolvedAlgo = Checksum.supportedAlgorithms.find(supportedAlgo => supportedAlgo === algo.toLowerCase())
```
Acconut, line 15: Same as before, let's use `Array#includes`.

lib/upload.js
```diff
@@ -531,6 +534,40 @@ class BaseUpload {
       return
     }
 
+    if (this.options.enableChecksum) {
+      const optionsReq = this._openRequest('OPTIONS', this.options.endpoint)
```
Acconut, line 538: Currently, we do not check extension support for any extension that is used by tus-js-client. We assume that the user knows that the server supports the extension if they enable it. I would be in favor of keeping in line with this style and against making an OPTIONS request here.

lib/upload.js
```diff
@@ -531,6 +534,40 @@ class BaseUpload {
...
+            return
+          }
+          const supportedAlgorithms = res.getHeader('Tus-Checksum-Algorithm')
+          if (!supportedAlgorithms) {
+            throw new Error(
+              'tus: checksum is enabled, but Tus-Checksum-Algorithm is not defined by tus server',
+            )
+          }
+          let Checksum
+          if (window) {
```
Acconut, line 556: We handle code difference between Node.js and Browsers APIs using dependency injections in two different `index.js` files. See https://github.com/tus/tus-js-client/blob/1db985ac2a1531184a26e3e989de4a2d2f90a823/lib/node/index.js#L6-L18 and https://github.com/tus/tus-js-client/blob/1db985ac2a1531184a26e3e989de4a2d2f90a823/lib/browser/index.js#L6-L18. Let's do this here as well to avoid these requires in here.

lib/upload.js
```diff
@@ -531,6 +534,40 @@ class BaseUpload {
...
+            // eslint-disable-next-line global-require
+            Checksum = require('./browser/extensions/checksum').default
+          } else {
+            // eslint-disable-next-line global-require
+            Checksum = require('./browser/extensions/checksum').default
+          }
+          this._checksum = new Checksum(
+            Checksum.supportedAlogrithms.find((supportedAlgo) => supportedAlgorithms.includes(supportedAlgo)),
```
Acconut, line 564: What does this check do? The combination of find and includes looks suspicious.

lib/upload.js
```diff
@@ -749,11 +786,25 @@ class BaseUpload {
...
+          if (this._checksum) {
+            value.arrayBuffer().then((chunkBuffer) => {
+              this._checksum
+                .getHexDigest(chunkBuffer)
+                .then((hash) => {
+                  req.setHeader('Upload-Checksum', `${this._checksum.algo} ${hash}`)
+                  this._emitProgress(this._offset, this._size)
+                  return this._sendRequest(req, value)
+                })
+                .catch((err) => log(err))
```
Acconut, line 799: This error should be bubbled up and not just logged.

lib/upload.js
```diff
@@ -749,11 +786,25 @@ class BaseUpload {
           req.setHeader('Upload-Length', this._size)
         }
 
-        if (value === null) {
+        if (this.options.enableChecksum) {
+          if (this._checksum) {
+            value.arrayBuffer().then((chunkBuffer) => {
```
Acconut, line 791: Could this entire if block be merged with the last else block? They seem very similar and would avoid some duplication.

Acconut (changes requested):
Thank you again for this incredible work. I left some comments to your code :)

test/spec/test-browser-specific.js
```diff
@@ -3,6 +3,7 @@
 const assertUrlStorage = require('./helpers/assertUrlStorage')
 const { TestHttpStack, waitableFunction, wait } = require('./helpers/utils')
 const tus = require('../..')
+const { default: Checksum } = require('../../lib.es5/browser/extensions/checksum')
```
Acconut, line 6: We should not require files from lib.es5 here, I think.

lib/upload.js
manohar27: I was trying to fetch the supported algorithms from the server and see if one of them matches one of the supported  algorithms on the client. If we're saying client and server have agreed upon an algorithm from the get go, then we can skip this OPTIONS call and let client just configure the algorithm in options.

lib/browser/extensions/checksum.js
manohar27: Done, added.

lib/upload.js
manohar27: Not sure how this might reduce the duplication as one is async. I've removed one outer if condition. Is that okay?

manohar27:
Thank you for the review. Please have a look @Acconut, I see there are conflicts. I'll try to address them, might need help

manohar27:
@Acconut please have another look at this

mahyarmirrashed:
Can I work on this?

Acconut:
Apologies, I wasn't aware that @manohar27 performed changes after my last review! 

@mahyarmirrashed Feel free to continue, but did you have anything specific in mind that needs to be changed here?

mahyarmirrashed:
@Acconut I have not taken a close look at the PR. Just wondering if there is a way that I can contribute/continue if needed. I might have a need for the checksum extension in about a month.

test/spec/test-browser-specific.js
```diff
@@ -787,4 +787,24 @@ describe('tus', () => {
       await assertUrlStorage(tus.defaultOptions.urlStorage)
     })
   })
+
+  describe('#Checksum', () => {
```
Acconut, line 791: We would also need a full test to ensure the checksum header is added to the actual requests. Such a test should be placed in `test-common`, which includes tests shared between browsers and Node.js.

Let me know if you need help with getting this running :)

lib/index.d.ts
```diff
@@ -129,3 +131,7 @@ export class DetailedError extends Error {
   originalResponse: HttpResponse
   causingError: Error
 }
+
+export interface Checksum {
```
Acconut, line 135: Once we have finalized the interface and options, we should also add documentation. But that can wait for now.

Acconut (changes requested):
Again, thanks very much for the work and apologies for losing track of this! I left a few comments if you or anybody else is interested in continuing this. It's on a very good track and we just need a few adjustments to get this ready.

mahyarmirrashed:
I'll try to do some work on this in the new year. By the way @Acconut , should there be any documentation changes with this? Something to let people know how to use this feature?

Acconut:
> should there be any documentation changes with this? Something to let people know how to use this feature?

Yes, once we have settled on how the feature's API should work, we can also address the documentation.

> I'll try to do some work on this in the new year.

Great! I am looking forward to your contributions :)

PR #458, opened 2022-09-07
Fix `Cannot call a namespace` error

aduh95:


Acconut:
Can you provide more context on when this error appears? I have never observed that

aduh95:
The code tries to call the namespace here: https://github.com/tus/tus-js-client/blob/6360e7845c0e3f81dec5e060b854e9491e863a37/lib/node/urlStorage.js#L116

That can obviously not work, as namespace imports always create a basic object (see https://tc39.es/ecma262/#sec-static-semantics-importentriesformodule, https://tc39.es/ecma262/#importentry-record, https://tc39.es/ecma262/#sec-getmodulenamespace, and https://tc39.es/ecma262/#sec-modulenamespacecreate), never a function.

Acconut (approved):
Ah, that all makes sense now. Thank you!

PR #459, opened 2022-09-08
require to use extensions in import specifiers

aduh95:
This fixes an error where bundlers would report a `ModuleNotFound` error because the extensions are missing:

https://codesandbox.io/s/nifty-fire-7luyeb?file=/package.json:252-350


.eslintrc
```diff
@@ -4,9 +4,10 @@
   ],
   "env": {
     "es6": true,
-    "node": true,
+    "node": false,
+    "jest": false,
     "browser": true,
```
Acconut, line 9: Can we also enable the browser env only for files in `lib/browser/`?

demos/nodejs/index.js
```diff
@@ -1,7 +1,9 @@
 /* eslint-disable no-console */
 
+'use strict'
+
 const fs = require('fs')
-const tus = require('../..')
+const tus = require('../..') // eslint-disable-line import/no-unresolved
```
Acconut, line 6: What is this eslint comment about? The files should be present once the build scripts has been run?

demos/browser/demo.js
```diff
@@ -1,6 +1,8 @@
 /* global tus */
 /* eslint-disable no-console, no-alert */
 
+'use strict'
```
Acconut, line 4: Why do some files get strict mode but not all?

Acconut (changes requested):

demos/nodejs/index.js
aduh95: Ah that might be because I didn't build before running the linter indeed.

demos/browser/demo.js
aduh95: See https://tc39.es/ecma262/#sec-strict-mode-code: 

> [Module code](https://tc39.es/ecma262/#sec-types-of-source-code) is always strict mode code.

That means, only non-ESM files need the `'use strict'` directive.

.eslintrc
aduh95: Good call. That's done, the linter does complain about the use of `window`, but nothing a `// eslint-disable-line` comment can't fix.

demos/browser/demo.js
Acconut: I see, that you!

Acconut (approved):
Thank you!

Acconut:
I just released 3.0.1 with these patches.

Issue #524, opened 2022-12-16
About `_onUploadUrlAvailable`


cjauvin:
**Question**

I have a use case where I need to know `upload.url` as soon as possible.. I tried and it seems I can use the `_onUploadUrlAvailable` option callback to retrieve it, but then as it's not part of the `UploadOptions` interface, it seems dangerous for me to use it. Is there any reason why this callback is not part of the option interface?

Acconut:
There is no specific reason for keeping it private. The new for this callback arose from internal demand, so we made it non-public by default. But I think there is nothing speaking against making it public now. Feel free to open a PR for this :)

Acconut:
Fixed in #543

PR #543, opened 2023-01-18
Expose the onUploadUrlAvailable callback option

cjauvin:
Following this conversation https://github.com/tus/tus-js-client/issues/524, I would like to propose to make the `onUploadUrlAvailable` callback officially and publicly available as an option.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -94,17 +94,23 @@ onShouldRetry: function (err, retryAttempt, options) {
     console.log("Error", err)
     console.log("Request", err.originalRequest)
     console.log("Response", err.originalResponse)
-    
+
     var status = err.originalResponse ? err.originalResponse.getStatus() : 0
     // Do not retry if the status is a 403.
     if (status === 403) {
       return false
     }
-    
+
     // For any other status code, we retry.
     return true
 }
 ```
+
+#### onUploadUrlAvailable
+
+*Default value:* `null`
+
+An optional function called once the upload URL is retrieved from the `Location` header of the initial creation POST's response. At this point, `upload.url` is guaranteed to be accessible and valid.
 
 #### headers
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -110,7 +110,7 @@ onShouldRetry: function (err, retryAttempt, options) {
 
 *Default value:* `null`
 
-An optional function called once the upload URL is retrieved from the `Location` header of the initial creation POST's response. At this point, `upload.url` is guaranteed to be accessible and valid.
+An optional function called once the upload URL is retrieved from the `Location` header of the initial creation POST's response, and also when an upload URL is confirmed using a HEAD request. At these points, `upload.url` is guaranteed to be accessible and valid.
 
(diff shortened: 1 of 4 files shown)


jean-kiepura:
👍 🥇

docs/api.md
```diff
@@ -94,18 +94,24 @@ onShouldRetry: function (err, retryAttempt, options) {
...
     // For any other status code, we retry.
     return true
 }
 ```
 
+#### onUploadUrlAvailable
+
+*Default value:* `null`
+
+An optional function called once the upload URL is retrieved from the `Location` header of the initial creation POST's response. At this point, `upload.url` is guaranteed to be accessible and valid.
```
Acconut, line 113: The callback is invoked in two cases: 1) A new upload has been created using the POST method, or 2) an upload URL was confirmed using a HEAD request. In addition, the callback might be invoked multiple times when multiple HEAD requests occur in a single upload due to connection interruptions.

I think these facts should be added to the documentation as well.

test/spec/test-common.js
```diff
@@ -304,6 +304,35 @@ describe('tus', () => {
...
+      req.respondWith({
+        status         : 204,
+        responseHeaders: {
+          'Upload-Offset': 11,
+          'Upload-Length': 11,
+        },
+      })
+
+      await options.onSuccess.toBeCalled
+      expect(options.onUploadUrlAvailable).toHaveBeenCalled()
```
Acconut, line 333: Can we move this assertion before `await options.onSuccess.toBeCalled` so we know that the callback is invoked directly after the HEAD response? Or does this asynchronous behavior interfere with this?

test/spec/test-common.js
```diff
@@ -304,6 +304,35 @@ describe('tus', () => {
       expect(options.onAfterResponse).toHaveBeenCalled()
     })
 
+    it('should invoke the onUploadUrlAvailable callback', async () => {
```
Acconut, line 307: This test covers the callback for HEAD responses. Can you also test its correct functionality with POST responses? It does not have to be a separate test but can also be added to another test covering POST requests.

Acconut (changes requested):
Great work! Thank you very much already!

docs/api.md
cjauvin: Sorry for the long delay! I have modified the documentation (as well as my understanding!) accordingly.

test/spec/test-common.js
cjauvin: I have decided to remove the entier extra test that I had added, and rather test that the callback is called in the two conditions that you mentioned, in two corresponding, already existing tests. However, as it seems that the callback is only called after `testStack.nextRequest()` resolves, the `expect` call must be done after it (which looks a bit strange, but is logical, as I verified explictly).

test/spec/test-common.js
cjauvin: See comment below.

Acconut (approved):
Thank you very much!

PR #556, opened 2023-02-23
Prevent potential size overflow and 0 length PATCH

mifi:
note: i'm not sure how to reproduce or even test this. and i'm not even sure if this is the correct solution, but I saw some code that I thought could be buggy.

alternatively we could throw an error if `sliceSize < 0` or `offset > this._size`, because that could indicate a logic issue.

see discussion in https://community.transloadit.com/t/how-to-abort-hanging-companion-uploads/16488/10

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -738,7 +738,7 @@ class BaseUpload {
     if ((end === Infinity || end > this._size) && !this.options.uploadLengthDeferred) {
       end = this._size
     }
-    
+
     const sliceSize = end - start
     if (sliceSize <= 0) { // done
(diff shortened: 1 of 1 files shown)


Acconut:
Tests are failing for your changes, which could indicate a problem. Also, in https://community.transloadit.com/t/how-to-abort-hanging-companion-uploads/16488/13 they mentioned that this behavior would occur when uploadSize was too big, so we should focus on fixing this.

cc @jdcalvin

Acconut:
Superseeded by https://github.com/tus/tus-js-client/pull/606.

Issue #579, opened 2023-04-13
HEAD request is not retried on intermittent server errors such as 502 Bad Gateway


nh2:
In

https://github.com/tus/tus-js-client/blob/01d3948c1a0b96adee545af72cdaabc5c5e979b1/lib/upload.js#L632-L676

the logic results in `HEAD` requests giving e.g. HTTP `502 Bad Gateway` not being retried; instead a new upload is started immediately. The user-defined `onShouldRetry()` is unexpectedly not called.

This is inconsistent with other parts of `tus-js-client`:

* `PATCH` requests are retried
* `POST` requests to create new uploads are retried on HTTP 502, because of
  https://github.com/tus/tus-js-client/blob/01d3948c1a0b96adee545af72cdaabc5c5e979b1/lib/upload.js#L592-L593
  where `_emitHttpError()` calls `_emitError()` which has the retry logic (`shouldRetry()` calling `onShouldRetry()`)

**To Reproduce**

1. Start an upload, e.g. to some TUS server with a reverse proxy such as nginx in between
2. Restart the upstream TUS server, so that HTTP `502 Bad Gateway` is generated by the reverse proxy
3. See error

Please try to reproduce the problem when uploading to our public tus instance at [https://tusd.tusdemo.net/files/](https://tusd.tusdemo.net/files/), if applicable to your situation. This helps us to determine whether the problem may be caused by the server component.

**Expected behavior**

* `HEAD` retries the same way as `PATCH`.
* `onShouldRetry()` should be called for `HEAD` the same way as for `PATCH`, so that the user can implement error handling.

nh2:
It also looks to me that https://github.com/tus/tus-js-client/blob/01d3948c1a0b96adee545af72cdaabc5c5e979b1/lib/upload.js#L656-L660

is too general.

For example, HTTP `429 Too Many Requests` for rate-limiting should allow the user to back off and retry after a short time, instead of creating a new upoad (thus creating even more requests).

It seems that a better approach is to list exactly the cases in which new uploads are created without calling the user's `onShouldRetry()`, instead of this broad handling.

Similar to https://github.com/tus/tus-js-client/blob/01d3948c1a0b96adee545af72cdaabc5c5e979b1/lib/upload.js#L1013 where cases are also listed explicitly.

PR #581, opened 2023-04-14
Fix HEAD not being retried for HTTP 500 category responses

nh2:
Tentative PR to help with #579.

This fixes the immediate issue of key HTTP 500 errors such as `502 Bad Gateway` not leading to retries.

But as described in https://github.com/tus/tus-js-client/issues/579#issuecomment-1508169053 I thikn that even more cases should lead to retry.

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -615,6 +615,11 @@ class BaseUpload {
         if (status === 423) {
           this._emitHttpError(req, res, 'tus: upload is currently locked; retry later')
           return
+        } else if (inStatusCategory(status, 500)) {
+          // Run retry logic if the server has an error, e.g. 502 Bad Gateway when
+          // proxied server is temporarily down. See issue #579.
+          this._emitHttpError(req, res, 'tus: resuming upload resulted in HTTP 500 category error')
+          return
         }
 
         if (inStatusCategory(status, 400)) {


lib/upload.js
```diff
@@ -615,6 +615,11 @@ class BaseUpload {
         if (status === 423) {
           this._emitHttpError(req, res, 'tus: upload is currently locked; retry later')
           return
+        } else if (inStatusCategory(status, 500)) {
```
Acconut, line 618: Let's not use an `else if` here but just an `if`. The `else` should not be necessary.

lib/upload.js
```diff
@@ -615,6 +615,11 @@ class BaseUpload {
...
+          // Run retry logic if the server has an error, e.g. 502 Bad Gateway when
+          // proxied server is temporarily down. See issue #579.
```
Acconut, line 620: Please mention that we do not want to clear `this._url` and create a new upload here.

Acconut (changes requested):
Thank you! This should fix the problem. Could you please add a case for our test suite?

lib/upload.js
nh2: Done.

lib/upload.js
nh2: Done.

Acconut:
Thanks for the updates. Only a test case would be nice now:
> Could you please add a case for our test suite?

Let me know if you need help with that.

PR #637, opened 2023-10-19
move default retry logic on to defaultOptions

### LINKED ISSUE #636: It's not possible to declare onShouldRetry and fallback to the default behaviour
**Is your feature request related to a problem? Please describe.**
I want to catch a specific server error and not retry, but now I lose online checking and the status code checking that happens by default.

**Describe the solution you'd like**
I would like to be able to detect and handle my error, then fallback to the default should retry implementation provided by the library.

**Describe alternatives you've considered**
I've tried to implement the same thing myself, but I don't want to rewrite all that code myself.

**Can you provide help with implementing this feature?**
One way to solve this would be to move the code [here](https://github.com/tus/tus-js-client/blob/64b58d0e5a10b434fafa6cd0d43d72e25f740b6e/lib/upload.js#L1029-L1030) into the defaultOptions as the default value for the `onShouldRetry` option. Then I could access the default function and call it inside my function.

**Additional context**
Maybe I should be doing error handling differently, please let me know if that's the case, I'm still learning about tus.

hahn-kev:
closes #636 

it was suggested that the default value of `onShouldRetry` should be set to `shouldRetry` but that won't work because `shouldRetry` calls `onShouldRetry`, so I opted to move the code at the bottom of `shouldRetry` into a function declared on `defaultOptions`. I wasn't quite sure what we wanted to do if a user explicitly set `onShouldRetry` to null. Currently the default logic would still be used, so that's what I preserved here, though it does look a bit awkward.

I wasn't able to run tests or execute a build locally. I kept getting a build error about a subdirectory or file already existing.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -33,7 +33,7 @@ interface UploadOptions {
   onSuccess?: (() => void) | null
   onError?: ((error: Error | DetailedError) => void) | null
   onShouldRetry?:
-    | ((error: Error | DetailedError, retryAttempt: number, options: UploadOptions) => boolean)
+    | ((error: DetailedError, retryAttempt: number, options: UploadOptions) => boolean)
     | null
   onUploadUrlAvailable?: (() => void) | null
 
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -23,7 +23,10 @@ const defaultOptions = {
   addRequestId: false,
   onBeforeRequest: null,
   onAfterResponse: null,
-  onShouldRetry: null,
+  onShouldRetry: (err, retryAttempt, options) => {
+    const status = err.originalResponse ? err.originalResponse.getStatus() : 0
+    return (!inStatusCategory(status, 400) || status === 409 || status === 423) && isOnline()
+  },
 
   chunkSize: Infinity,
   retryDelays: [0, 1000, 3000, 5000],
@@ -1026,8 +1029,7 @@ function shouldRetry(err, retryAttempt, options) {
     return options.onShouldRetry(err, retryAttempt, options)
   }
 
-  const status = err.originalResponse ? err.originalResponse.getStatus() : 0
-  return (!inStatusCategory(status, 400) || status === 409 || status === 423) && isOnline()
+  return defaultOptions.onShouldRetry(err, retryAttempt, options);
 }
 
 /**


hahn-kev:
I'm not sure what you mean. If we provide a default value and a user also provides a value the default will be overwritten, then the exact same thing that happens today does. Today if a user provides a function then the default functionality does not run. That hasn't changed.

Also, I didn't expose `shouldRetry` at all, I just moved some functionality inside of it into the default function. The only thing that's changed from the outside is that if you call `defaultOptions.onShouldRetry` you will get the default functionality if you didn't provide that option.

lib/index.d.ts
```diff
@@ -33,7 +33,7 @@ interface UploadOptions {
   onSuccess?: (() => void) | null
   onError?: ((error: Error | DetailedError) => void) | null
   onShouldRetry?:
-    | ((error: Error | DetailedError, retryAttempt: number, options: UploadOptions) => boolean)
+    | ((error: DetailedError, retryAttempt: number, options: UploadOptions) => boolean)
```
Acconut, line 36: Please keep `Error` here as the type is not always DetailedError: https://github.com/tus/tus-js-client/pull/287#issuecomment-945199333

lib/upload.js
```diff
@@ -23,7 +23,10 @@ const defaultOptions = {
   addRequestId: false,
   onBeforeRequest: null,
   onAfterResponse: null,
-  onShouldRetry: null,
+  onShouldRetry: (err) => {
```
Acconut, line 26: Please don't use an anonymous function here but put it in a function called `defaultOnShouldRetry`. This keep the definition of `defaultOptions` cleaner.

Acconut (changes requested):
You are right, this PR does not present a breaking change for users.

I have two comments on the code. Could you also please have a look at the lint issues in `upload.js`: https://github.com/tus/tus-js-client/actions/runs/6573176315/job/17950213388?pr=637#step:7:43

lib/index.d.ts
hahn-kev: But that's not revevant here.  In `shouldRetry` calls `onShouldRetry` if `err.originalRequest != null` which means that it can't just be an error.

lib/upload.js
hahn-kev: done

hahn-kev:
Alright, I've pushed changes and left a comment about the type change.

When I run lint on my machine it gives a bunch of errors and warnings about line endings and files I didn't change, but I think the lint error is fixed.

lib/index.d.ts
Acconut: Good point, then let's change it 👍

Acconut (approved):
Thank you for your help!

PR #670, opened 2024-02-21
fix: types incorrectly flag values as potentially null

aduh95:
AFAICT, every time we call `.terminate()` or `.fingerprint()`, we pass an `options` which cannot be nullish.
The `defaultValues` also have some always defined values, IMO it makes sense to report them as defined to avoid downstream code to have to use non-null-assertions.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -13,7 +13,7 @@ export class Upload {
   options: UploadOptions
   url: string | null
 
-  static terminate(url: string, options: UploadOptions): Promise<void>
+  static terminate(url: string, options?: UploadOptions): Promise<void>
   start(): void
   abort(shouldTerminate?: boolean): Promise<void>
   findPreviousUploads(): Promise<PreviousUpload[]>
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -2,12 +2,8 @@
 
 export const isSupported: boolean
 export const canStoreURLs: boolean
-export const defaultOptions: UploadOptions & Required<Pick<UploadOptions, 
-| 'httpStack'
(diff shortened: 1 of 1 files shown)


Acconut:
Sorry, I don't fully understand. Can you share exemplary code and the errors that you are running into?

aduh95:
Well before this PR, the types are "wrong" (or rather imprecise), e.g. `tus.defaultOptions.fingerprint` is typed as potentially nullish; after this PR, the types are more "correct", e.g. `tus.defaultOptions.fingerprint` is typed as not nullish.

Acconut:
Ok, that makes sense. I get the change for `defaultOptions` now, thank you. But why are you making the `options` argument for `terminate` non-nullish now? It can be `null`: https://github.com/tus/tus-js-client/blob/95ca56674e821f928a583682fba8c9ecd8dff14b/lib/upload.js#L118

Am I missing something else?

aduh95:
> why are you making the `options` argument for `terminate` non-nullish now? It can be `null`

Maybe this one is wrong, not sure. Let's remove this change.

lib/index.d.ts
```diff
@@ -12,7 +13,7 @@ export class Upload {
   options: UploadOptions
   url: string | null
 
-  static terminate(url: string, options?: UploadOptions): Promise<void>
+  static terminate(url: string, options: UploadOptions): Promise<void>
```
aduh95, line 16:
```suggestion
  static terminate(url: string, options?: UploadOptions): Promise<void>
```

Acconut (approved):
Alright, then this looks good, thank you!

PR #675, opened 2024-03-15
Performance optimization of StreamSource

magoogli:
At the moment when a call to 'slice(start, end)' occurs on StreamSource, if buffers in memory all data from wherever the current position in the stream is up to end. 

This is a serious problem if the stream has never been read from and you call slice with a start and end position near the end of the stream. E.g if I have a 2 gigabyte file on disk that I am reading via a Web ReadableStream, and am using that as the source of my data for the StreamSource, and I ask for the last megabyte slice at the end of the StreamSource it will currently buffer the entire 2 gigabytes into a memory buffer, before returning the last megabyte!

Obviously with a stream you will always have to read the entire 2 gigabytes to get to the end, but buffering the data in memory is much more problematic and wasteful than just reading through the stream.


magoogli:
As to your question about real world use - Yes, I am creating a new tus.Upload object. When dealing with a standard javscript File object the scenario I described above works without issue ( as with a file you can get a slice without having to read any data prior to 'start' ). I then started playing around with the idea of on the fly compression using the Compressions Streams API and starting observing the massive slowdown in the scenario above when using a stream. After some debugging I found out that it was all the concatenation and buffering in memory which was causing the issue.

magoogli:
@Acconut I fixed the bug in my code and files are uploaded correctly now. The optimization definitely helps for the use case I described ( in a real world application we are developing ). Would you like any further clarification or code changes?

Acconut:
That change itself looks good and also makes sense to me. Before we can merge this, we should also add a test case for this. In https://github.com/tus/tus-js-client/blob/f410679e8ae4b3821ffb919a5cfe7f04acd1528d/test/spec/test-browser-specific.js/#L251, we have test cases specific to readers in browsers. There a test case can be added, where a stream is supplied for resuming an upload with a non-zero offset. Seeking to this offset should result in multiple calls to the reader. In the end, an assertion should take place confirming that the data in the PATCH request is correct. Could you look into this?

magoogli:
Hi @Acconut thanks for checking out my updates. I will try and push a test for the changes in the next few days.

Acconut:
Apologies for dropping the ball on this, @magoogli. I just merged the latest changes from `main` into your PR while making sure the stream handling logic and test are preserved properly. If CI is happy, I'll merge this. Thank you very much your effort on this!

Acconut (approved):

magoogli:
> Apologies for dropping the ball on this, @magoogli. I just merged the latest changes from `main` into your PR while making sure the stream handling logic and test are preserved properly. If CI is happy, I'll merge this. Thank you very much your effort on this!

No problem at all, and thanks for merging it in!

Issue #679, opened 2024-03-27
Decrease browser bundle size


mb21:
Looking at the bundle files for `lib.esm` for example:

<img width="747" alt="image" src="https://github.com/tus/tus-js-client/assets/157497/cce7da5c-e1db-4c32-b93d-bb47c2a394bf">

(this was generated in an Astro.js project with `rollup-plugin-visualizer`).

I was especially curious why `lib.esm/browser/fileReader.js` is so big even though the [source file](https://github.com/tus/tus-js-client/blob/main/lib/browser/fileReader.js) is only a few lines. Running `yarn run build-transpile-esm` results in a 14kb file. The problem is that babel includes all the following functions:

```
_typeof
_regeneratorRuntim
asyncGeneratorStep
_asyncToGenerator
_classCallCheck
_defineProperties
_createClass
_toPropertyKey
_toPrimitive
```

Most of which is not needed for more or less modern browsers anymore.

I'm not  an expert in babel and npm package bundling, but AFAIK it's nowadays best practice for a JS library to kinda just ship the source, so that the bundler in the project that consumes the library can optimize for the browsers they want to support and tree-shake away the functions not used?

Acconut:
Thank you for looking into this!



> I'm not an expert in babel and npm package bundling, but AFAIK it's nowadays best practice for a JS library to kinda just ship the source, so that the bundler in the project that consumes the library can optimize for the browsers they want to support and tree-shake away the functions not used?

I agree, we should get rid of the polyfill for async/await and generators. We can assume support for those features in the tooling by now.

PR #681, opened 2024-04-11
Decrease browser bundle size

mb21:
by optimizing for more modern browsers.

see #https://github.com/tus/tus-js-client/issues/679


Acconut:
Thank you for the investigation and this PR! 

> Optimizing for more modern browsers by removing [forceAllTransforms](https://babeljs.io/docs/babel-preset-env#forcealltransforms), which was added in https://github.com/tus/tus-js-client/pull/192 by @goto-bus-stop (as part of ESM support), although the PR description seems to imply that forceAllTransforms wasn't intended. (this makes the esm browser bundle go from from 97KB down to 47KB)

Good point! I am no expert at transpiling and bundling, which seems to be a whole science on its own. Nowadays we can raise our requirements on the ECMAScript support by the user's bundler and browsers. Although this has to happen in a major release to avoid breaking changes in a minor release. That being said, I have recently worked a bit on moving tus-js-client to TypeScript (#683), which was also a good opportunity to raise the compilation target.

> 3\. Removing the js-base64 package in favour of [btoa](https://stackoverflow.com/a/247261/214446) (saves 9.6KB)

This dependency was added to support tus-js-client inside React Native environment, [where `btoa` is not available](https://stackoverflow.com/questions/42829838/react-native-atob-btoa-not-working-without-remote-js-debugging). See the pull request https://github.com/tus/tus-js-client/pull/132 and especially the commit https://github.com/tus/tus-js-client/pull/132/commits/0c72e860e96e4cb29ea82cfd9bb06700a9c3adb7. Although the [React Native JS runtime](https://reactnative.dev/docs/javascript-environment) looks similar to the environment in a browser, React Native apps cannot utilize all Web APIs and `btoa` is not provided. Hence we need a polyfill for this situation.

That being said, ~10KB for base64 encoding seems a bit too much and there are two possible approaches I can think of
- try to only import the encoding function in the hopes that tree shaking is able to remove the rest of the `js-base64` package: `import {encode} from 'js-base64'`
- write our own base64 encoding function instead of relying on packages or environment support. That should be doable in less than 10KB.

Ideally, we would only need a polyfill for React Native, but I am not sure how that would be possible.

> 2\. Removing the [url-parse](https://www.npmjs.com/package/url-parse) package in favour of Node/browser-native `new Url` (saves 16KB)

Similar story as before. [`URL` is not available in React Native](https://github.com/facebook/react-native/issues/16434) and thus we needed a polyfill: https://github.com/tus/tus-js-client/commit/53a6ac3c50cf0106b5ae1d350451a0859e626624.

An advantage of using those packages instead of using environment-specific APIs is that we get more consistent behavior, which is not influenced by subtle different between the implementations.

> 4\. Add info to `package.json` for https://volta.sh, so that you can easily install the old yarn version used in this project (feel free to remove that commit if you think it doesn't belong).

I don't know about Volta, but such a change should be in a separate PR and not part of this one. That being said, maybe it's worth considering switching back to NPM, which does not have different versions as Yarn appears to have.

Acconut:
I am hesitant with dynamic imports as I fear that they make the entire setup more complex and brittle. A separate build would also be possible, but I would prefer we first explore solutions to reduce the size of the base64 encoder, either by adjusting the import statement or pulling the code into tus-js-client directly.

mb21:
Yeah, you could use [this implementation](https://github.com/MaxArt2501/base64-js/blob/master/base64.js#L33). But polyfilling `URL` is probably going to be bigger.

mb21:
Ah, I've never done React Native development, but [`react-native-url-polyfill` says](https://www.npmjs.com/package/react-native-url-polyfill):

> React Native does include a polyfill for URL, but this polyfill is homemade — in order to keep it light-weight — and was initially created to handle specific use cases.

mb21:
Either way, I would avoid publishing the bundled files to npm. (See e.g. this [StackOverflow question](https://stackoverflow.com/questions/69514389/best-practice-for-npm-package-with-es6-modules-bundle-or-not)). Not sure how the TypeScript conversion is going, but it's possible that using `tsc` instead of `esbuild` will be enough.

Acconut:
> Either way, I would avoid publishing the bundled files to npm.

We currently include a bundled version alongside the original source code and a transpiled version. The latter two are referenced in the package.json, so the bundler can decide whatever variant it prefers using. The bundled version is not used in such cases. It is only included, so that people can download the bundle from NPM or use it via services like [jsDelivr](https://www.jsdelivr.com/package/npm/tus-js-client). Under such circumstances, I don't think it's harmful to include a bundled version in the package. The StackOverflow question you linked seems to refer to package which only include a bundled version, which is not the case for tus-js-client (and we intend to keep it that way).

Acconut:
> 2\. Removing the [url-parse](https://www.npmjs.com/package/url-parse) package in favour of Node/browser-native `new Url` (saves 16KB)

An alternative to `url-parse` could be `url-toolkit`. I haven't test it yet and its documentation mentions some differences to `URL`, but it's designed to be lightweight and is used in the popular hls.js library: https://github.com/tjenkinson/url-toolkit Maybe the bundle weight can be reduced that way.

PR #683, opened 2024-04-16
Move to TypeScript

Acconut:
This PR moves the libraries' code base to TypeScript. The benefit is that we have type-checked source code and also don't have to manually maintain a type definition in `index.d.ts`. The source code is transpiled into separate CommonJS and ESM variants. The PR does not change the dual-package setup of tus-js-client.

TODOs:
- [x] Merge latest main and fix conflicts
- [x] Update PR description
- [x] Look for missing types
- [x] Release last non-TypeScript release before merging this


Acconut:
@Murderlon You mentioned that dual publishing a package containing CommonJS and ESM can lead to unexpected problems. Are there some links or resources digging more into this? Are you referring to what Node.js calls the [dual package hazard](https://nodejs.org/api/packages.html#dual-package-hazard)? We have dual published this package for four years (since https://github.com/tus/tus-js-client/commit/bdb614438378c938444456941496da534414df6b) and didn't run into any problems and no major issues where reported to us.

By completely ditching CommonJS, I am worrying that we leave some application behind which are still using it and have a hard time upgrading to ESM (I think to recall that our website tus.io also still compiles to CommonJS, but I might get that wrong). Is this concern unjustified nowadays?

Murderlon:
> You mentioned that dual publishing a package containing CommonJS and ESM can lead to unexpected problems. Are there some links or resources digging more into this?

This from my experience + some other prominent maintainers I talk with. All my issues in the past with a package were from dual publishing, not ESM. Examples include Preact itself and even the package [dual-publish](https://github.com/ai/dual-publish), which is supposed to solve it.

It is possible and since we already have it working it's not really a problem. But at some point we should make the jump, ESM is the future.

> By completely ditching CommonJS, I am worrying that we leave some application behind which are still using it and have a hard time upgrading to ESM

Technically everyone can use it, even in a CommonJS setup:

```js
const tusPromise = import('tus-js-client')

module.exports = tusPromise.then((tus) => tus.default)
```

It's not the ideal developer experience, but it's important to note that tus-js-client is close to done for tus protocol v1. It's stable and it's totally fine to stay on the previous major version if you must for a while. And even without ditching CommonJS, this would be a major version anyway, and I think it's better to have less major versions (rather than major version for TS then later again for ESM).

---

Some tools which are importent to check:

- https://github.com/wooorm/package-exports (gives errors and warnings based on our `package.json`)
- https://arethetypeswrong.github.io/?p=tus-js-client%404.1.0
- [Information for users who now have to deal with an ESM package](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c).

Murderlon:
Some more warnings:
https://publint.dev/tus-js-client@4.1.0

PR #689, opened 2024-05-14
Provide last response to `onSuccess` callback

netdown:
Support using customized responses as introduced in tus/tus-node-server#615


netdown:
Indeed I wasn't cautious enough. What if the argument is only passed for post and patch requests' events? Or maybe like `{ requestType: 'PATCH'|'POST'|'HEAD', response: HttpResponse }`;

My goal is being able to access response data in Uppy's `complete` event using `uppy.getFiles()`. Unfortunately neither headers nor status code (actually the response object contains a status code, but the value is hard coded to 200) are returned using tus, so I know nothing but the upload url.

Acconut:
Have you tried the [onAfterResponse](https://github.com/tus/tus-js-client/blob/main/docs/api.md#onafterresponse) callback? Did that not work for you? It should also be usable from Uppy.

netdown:
It feels really hacky that I'd need to decide which response is a success response and store them in a custom state, then at the complete handler, make sure my state is actually in sync with uppy's state, find the appropriate response for each file, and make sure uppy processing haven't failed for the responses I assumed to be ok. Seems too complicated and prone to bugs while it could be solved by passing the response. Are you sure the object signature I presented is not suitable?

Acconut:
> Are you sure the object signature I presented is not suitable?

No, I am not very sure and cannot guarantee that my opinion doesn't change :) I am trying to considering the bigger picture here and how we can improve the overall approach. The problem is that in tus, there is no standardized way of sending information from the server back to the client as tus focuses on uploads where the client sends data to the server. This has lead to the current situation where additional data from the server is added to PATCH responses for clients to retrieve. But this is not ideal because the information in these responses cannot be re-retrieved if the clients needs to fetch it again. So overall, I would like to have a more holistic approach for sending data from the server, which solves this problem in a proper way without shortcomings. Some discussion on this has happened in https://github.com/tus/tus-resumable-upload-protocol/issues/190.

Therefore, I would like to reserve a future argument for `onSuccess` to be whatever approach we come up with there instead of using it right now for the last response right now, which might not make sense in the future. Is that understandable?

Acconut:
I can understand that an extra round trip for fetching the Content-Location resource is not desired in your case. You will always have the ability to add custom headers to responses and retrieve them on the client side to implement custom logic.

> Additionally, I'm not sure requerying some data about the uploaded file is not out of tus territory. The upload has been completed, why bother post-upload logic?

That's definitely a question worth asking and my opinioned in the past year has developed to be that tus implementations should assist you with postprocessing in some way as this makes file handling in applications more straightforward, robust and seamsless. Of course, that doesn't mean that we have to offer a full solution for postprocessing files. But these questions come up very often and I would prefer if we can offer some answer. Using them is then up to each person on their own.

netdown:
Sorry for the late reply. I have modified the signature and added a test.

docs/api.md
```diff
@@ -61,7 +61,14 @@ An optional function that will be called each time a `PATCH` has been successful
 
 _Default value:_ `null`
 
-An optional function called when the upload finished successfully.
+An optional function called when the upload finished successfully. The argument will be an object with information about the event. Its `lastResponse` key will contain the lastly received HttpResponse, which can be used to retrieve additional data from the server. For example:
```
Acconut, line 64:
```suggestion
An optional function called when the upload finished successfully.

The argument will be an object with information about the completed upload. Its `lastResponse` property contains lastly received [`HttpResponse`](#httpstack), which can be used to retrieve additional data from the server. Be aware that this is usually a response to a `PATCH` request, but it might also be a response for a `POST` request (if `uploadDataDuringCreation` is enabled or an empty file is uploaded) or `HEAD` request (if an already completed upload is resumed).
```

Some more details about the response.

docs/api.md
```diff
@@ -61,7 +61,14 @@ An optional function that will be called each time a `PATCH` has been successful
...
+
+```js
+onSuccess: function (payload) {
+	const { lastResponse } = payload;
+    console.log("Status code", lastResponse.getStatus())
```
Acconut, line 69: A bit more useful example

```suggestion
	  const { lastResponse } = payload
	  // Server can include details in the Upload-Info header, for example.
    console.log("Important information", lastResponse.getHeader('Upload-Info'))
```

lib/upload.js
```diff
@@ -520,17 +520,18 @@ class BaseUpload {
   /**
    * Publishes notification if the upload has been successfully completed.
    *
+   * @param {object} res Last HTTP response.
    * @api private
    */
-  _emitSuccess() {
+  _emitSuccess(res) {
```
Acconut, line 526: Please rename this argument to `lastResponse`.

Acconut (changes requested):
Thank you for updating this PR! I proposed a few more changes. Could you also look into the merge conflict? If these two points get resolved, we can merge this.

netdown:
All done.

Acconut (approved):
Thank you very much for your contribution!

PR #690, opened 2024-05-27
Switch from ESLint & Prettier to Biome

Acconut:
This PR replaces ESLint and Prettier with Biome, resulting in a less complex configuration file for linting and formatting. It is based on Biome's recommended rules plus some rules that were suggested by Biome's migration script. Some changes to our source code was necessary to please the new tool, but I think they are fair overall.


biome.json
```diff
@@ -0,0 +1,88 @@
...
+  "linter": {
+    "enabled": true,
+    "rules": {
+      "style": {
+        "noParameterAssign": "off",
+        "noDefaultExport": "warn",
+        "useCollapsedElseIf": "error"
+      },
+      "suspicious": {
+        "noExplicitAny": "off",
```
kvz, line 24: `unknown` is better than `any`. If you _really_ must use `any`, it may be preferable to disable the rule at line level with an explanation, as `any`s have a way of sneaking up on you. A deep `any` may sweep type errors under the rug high up the dependency chain, without any signs to the developer, until you hover over the variable and it turns out to be an `any`

biome.json
```diff
@@ -0,0 +1,88 @@
...
+    "attributePosition": "auto"
+  },
+  "organizeImports": {
+    "enabled": true
+  },
+  "linter": {
+    "enabled": true,
+    "rules": {
+      "style": {
+        "noParameterAssign": "off",
```
kvz, line 19: Should this be a `warn`? Parameter reassignment could lead to issues

lib/index.test-d.ts
```diff
@@ -21,14 +21,14 @@ const upload = new tus.Upload(file, {
...
+    console.log(bytesSent, bytesTotal, `${percentage}%`)
   },
-  onChunkComplete: (chunkSize: number, bytesAccepted: number) => {},
+  onChunkComplete: (_chunkSize: number, _bytesAccepted: number) => {},
   onSuccess: () => {
     console.log('Download from %s complete', upload.url)
   },
   onError: (error: Error | DetailedError) => {
-    console.log('Failed because: ' + error)
+    console.log(`Failed because: ${error}`)
```
kvz, line 31: Unrelated but it caught my eye
```suggestion
    console.error(`Failed because: ${error}`)
```

test/spec/test-parallel-uploads.js
```diff
@@ -1,5 +1,3 @@
-'use strict'
```
kvz, line 1: Do you know what the reasoning is behind this? Just curious

I thought you could remove for ts, if tsconfig had strict enabled, otherwise it still made sense, but I could be wrong.

kvz (commented):
Looking sensible! Just a few nits

Murderlon:
LGTM

test/spec/test-parallel-uploads.js
Acconut: Biome recommended removing this because of its [`noRedundantUseStrict` rule](https://biomejs.dev/linter/rules/no-redundant-use-strict/) that says that `'use strict'` is not necessary for `.js` files, when the nearest `package.json` includes `type: module`, which is the case for tus-js-client. Therefore, I removed it. However, I am also not sure if the tools we use for building and bundling will still use strict mode for those files. But I am willing to test it since those are only the test files and thus not directly delivered to the package users.

biome.json
Acconut: Good idea! I would like to fully enable this rule, but we violate it at many places and I prefer tackling them in a separate PR.

biome.json
Acconut: In the main branch we currently only use TypeScript in our `index.d.ts` definition files and I would like to keep it as unmodified as possible in this PR. Thus I will turn this into `warn` as well. With the TypeScript migration in #683, the entire types will be overhauled anyways. Thanks for bringing this up!

PR #701, opened 2024-07-09
Add metadata into parallel uploads.

loren-m-crawford:
👋  This PR is more of a proposal to allow for parallel uploads to keep the metadata from the original request. My team has found that maintaining metadata in the parallel upload requests would benefit our use case.

The PR does two main things:
1.  Adds an option `retainMetadataForParallelUploads` that is defaulted to false. This option would be for our use case (or anyone else's) that would allow the retention of metadata in partial uploads, too.
2. Checks the `retainMetadataForParallelUploads` option during the parallel upload logic. If it's set to true, then we ensure the metadata is maintained, if it is false (default behavior), we remove it.

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -331,7 +331,8 @@ class BaseUpload {
               parallelUploads: 1,
               // Reset this option as we are not doing a parallel upload.
               parallelUploadBoundaries: null,
-              metadata: {},
+              metadata: { ...this.options.metadata
+              },
               // Add the header to indicate the this is a partial upload.
               headers: {
                 ...this.options.headers,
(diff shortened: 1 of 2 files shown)


Acconut:
Thanks for opening this PR! You're right that tus-js-client currently does not offer a way to set metadata for partial uploads. This is not ideal, especially since we also mention that metadata can be used to identify the users corresponding to an upload. I wonder if there are also use case where one wants to use different meta data for partial uploads versus the final upload. In such cases, an option to retain the meta data would not be sufficient and we would need an option to specify custom metadata for the partial uploads.

What do you think?

loren-m-crawford:
@Acconut 👋 You make a good point. Perhaps instead of maintaining the metadata from the original request, there can be an additional option for parallel upload metadata that can either be the same as the original request metadata or be more tailored. Thinking maybe something like: https://github.com/loren-m-crawford/tus-js-client/pull/2/files
What do you think?

Acconut:
Superseeded by https://github.com/tus/tus-js-client/pull/703.

PR #703, opened 2024-07-16
Add `metadataForPartialUploads` option to set metadata for partial uploads only

loren-m-crawford:
This is an alternative option to this [PR](https://github.com/tus/tus-js-client/pull/701) which proposes having the option to retain metadata on partial upload requests. Instead of maintaining the metadata from the original request, this PR adds the option for `partialMetadata`,  which would allow folks to set metadata only for partial requests. This could be the same or different from the original metadata, which allows for more flexibility.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -166,14 +166,14 @@ metadata: {
 }
 ```
 
-#### partialMetadata
+#### metadataForPartialUploads
 
 _Default value:_ `{}`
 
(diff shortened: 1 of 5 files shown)


Acconut (approved):

Acconut:
I'm deeply sorry for not reviewing this earlier! I just had a look at it and polished the implementation, tests and docs. Thank you for working on this PR!

PR #709, opened 2024-08-23
Only pass strings to `HttpRequest#setHeader`

DraganDanicic:
- Numbers passed as header-values are casted to string type.
- Tests adjusted in a way that expected header values are now strings.

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -581,7 +581,7 @@ class BaseUpload {
     if (this.options.uploadLengthDeferred) {
       req.setHeader('Upload-Defer-Length', '1')
     } else {
-      req.setHeader('Upload-Length',`${this._size}`))
+      req.setHeader('Upload-Length', `${this._size}`)
     }
 
     // Add metadata if values have been added
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -155,23 +155,6 @@ class BaseUpload {
       })
   }
 
-  static convertToStringHeaderValue(val) {
-    if (val === null || val === undefined) {
-      return ''
-    } else if (Array.isArray(val)) {
-      return val.join(',')
-    } else if (
-      typeof val === 'string' ||
-      typeof val === 'number' ||
-      typeof val === 'boolean' ||
-      typeof val === 'symbol'
-    ) {
-      return val.toString()
-    } else if (typeof val === 'object') {
-      return Object.values(val).join(',')
-    }
-  }
-
   findPreviousUploads() {
     return this.options
       .fingerprint(this.file, this.options)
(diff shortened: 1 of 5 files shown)


Acconut:
Thank you for the PR! All of these should be addressed once we get https://github.com/tus/tus-js-client/pull/683 merged, which moves the code base to TypeScript. While working on it, I also discovered that we call setHeader sometimes with a number, which is not ideal. But we can also get this PR faster merged without waiting on the TypeScript PR.

lib/upload.js
```diff
@@ -973,7 +990,7 @@ function openRequest(method, url, options) {
   const headers = options.headers || {}
 
   for (const [name, value] of Object.entries(headers)) {
-    req.setHeader(name, value)
+    req.setHeader(name, BaseUpload.convertToStringHeaderValue(value))
```
Acconut, line 993: I don't think we should magically convert non-strings into strings here, especially since this is technically a breaking change. It's the user responsibility to pass proper values in here that match what the documentation mentions.

We can log a warning if the value is not a string and then convert the warning into an error in the next major release.

lib/upload.js
```diff
@@ -155,6 +155,23 @@ class BaseUpload {
       })
   }
 
+  static convertToStringHeaderValue(val) {
```
Acconut, line 158: The helper method is a good idea, but not really needed. I would like to avoid a large helper method like this that magically converts everything somehow into a string and instead use dedicated methods where we need to convert things into string.

Acconut (changes requested):

DraganDanicic:
> Thank you for the PR! All of these should be addressed once we get #683 merged, which moves the code base to TypeScript. While working on it, I also discovered that we call setHeader sometimes with a number, which is not ideal. But we can also get this PR faster merged without waiting on the TypeScript PR.

I agree on your points. Shall we then go with explicit conversion at-the-spot (this ->`${this._size}`), so we can patch the holes for the time being, until we get TS coverage. Would this present enough value for you at this stage?

Acconut:
> I agree on your points. Shall we then go with explicit conversion at-the-spot (this ->`${this._size}`), so we can patch the holes for the time being, until we get TS coverage. Would this present enough value for you at this stage?

Yes, that would be my preferred solution.

DraganDanicic:
> > I agree on your points. Shall we then go with explicit conversion at-the-spot (this ->`${this._size}`), so we can patch the holes for the time being, until we get TS coverage. Would this present enough value for you at this stage?
> 
> Yes, that would be my preferred solution.

I have adjusted the code, and it should now meet the requirements. Please, have a look.

Acconut (approved):
Thank you for this PR!

PR #725, opened 2024-09-17
uuid.js use native randomUUID() if available

oittaa:
https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID

All the major browsers support it:

https://caniuse.com/mdn-api_crypto_randomuuid

### DIFF
--- a/lib/uuid.js
+++ b/lib/uuid.js
@@ -11,6 +11,9 @@
  * @return {string} The generate UUID
  */
 export default function uuid() {
+  if (typeof(self.crypto.randomUUID) === typeof(Function)) {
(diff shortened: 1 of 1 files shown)


Acconut:
Thank you for the PR. Could you explain what benefit the native function offers?

Acconut:
Thank you for the additional details, I agree that using native implementations is preferable whenever possible. tus-js-client not only supports browser but also run in Node.js and mobile environments, such as React Native. So we also have to take these environments and their capabilities into considerations when thinking about dropping these internal implementations. Node.js also implements parts of the Web Crypto API as far as I know, but I'm not sure about React Native.

PR #727, opened 2024-09-26
react native file argument details in api.md

h3knix:
similar to the other platform explanations, react native specific implementation notes with links added


h3knix:
it took me a moment to figure out what the react native platform specific file argument was; so I figured that some additional information on the api doc itself would help others

Acconut (approved):
Thank you very much! That's a good improvement!

PR #734, opened 2024-11-02
Move to TypeScript: Upload, StreamSource

holloway:
Adding some types for browser StreamSource


Acconut (approved):
Thank you very much for your help! I pushed a few additional commits to make the test suite pass. Notably, the `bytesTotal` argument for progress events can be `null`. This was previously not properly documented and TypeScript helped discover this 🙌

PR #744, opened 2025-01-13
Set Upload-Length when appropriate

### LINKED ISSUE #182: Upload against a deferred resource does not work without options.uploadLengthDeferred
**Describe the bug**
If a remote resource is deferred and the server reports the "Upload-Defer-Length: 1" header,
tus-js-client is not able to upload against the resource without options.uploadLengthDeferred.
The protocol does not enfoce this. Upload-Length can be sent with the very next PATCH request.

**To Reproduce**
Create a deferred resource.
```bash
curl -v -X POST -H 'Tus-Resumable: 1.0.0' -H 'Upload-Defer-Length: 1' https://master.tus.io/files
```
Try to upload against the resource without options.uploadLengthDeferred
Error: "tus: invalid or missing length value"

**Expected behavior**
It should be possible to upload against the deferred resource without options.uploadLengthDeferred

**Setup details**

Vija02:
This PR fixes issue #182 where user cannot upload against a deferred resource.

The 2 main changes are:
1. Bypass the `Upload-Length` header check if `Upload-Defer-Length` exists in the HEAD response
2. Sets the `Upload-Length` header in the next request if `options.uploadLengthDeferred` is false - because we would know the length straight away

The code is mostly taken from the relevant issue, so credits goes there. But I've also modified it slightly to simplify the logic.

Closes #182

### DIFF
--- a/lib/upload.js
+++ b/lib/upload.js
@@ -108,7 +108,10 @@ class BaseUpload {
 
     // An array of upload URLs which are used for uploading the different
     // parts, if the parallelUploads option is used.
-    this._parallelUploadUrls = null
+    this._parallelUploadUrls = null    
+
+    // The remote upload resource's length is deferred
+    this._deferred = false
   }
 
   /**
@@ -586,6 +589,7 @@ class BaseUpload {
 
     if (this.options.uploadLengthDeferred) {
       req.setHeader('Upload-Defer-Length', '1')
+      this._deferred = true
     } else {
       req.setHeader('Upload-Length', `${this._size}`)
     }
@@ -704,9 +708,13 @@ class BaseUpload {
           return
         }
 
+        const deferLength = Number.parseInt(res.getHeader("Upload-Defer-Length"), 10)
+        this._deferred = deferLength === 1
+
         const length = Number.parseInt(res.getHeader('Upload-Length'), 10)
         if (
           Number.isNaN(length) &&
+          !this._deferred &&
           !this.options.uploadLengthDeferred &&
           this.options.protocol === PROTOCOL_TUS_V1
         ) {
@@ -821,9 +829,10 @@ class BaseUpload {
       // If the upload length is deferred, the upload size was not specified during
       // upload creation. So, if the file reader is done reading, we know the total
       // upload size and can tell the tus server.
-      if (this.options.uploadLengthDeferred && done) {
+      if (this._deferred && (!this.options.uploadLengthDeferred || done)) {
         this._size = this._offset + valueSize
         req.setHeader('Upload-Length', `${this._size}`)
+        this._deferred = false
       }
 
       // The specified uploadSize might not match the actual amount of data that a source


Acconut (changes requested):
Thank you for this PR! Overall, taking the information about deferred length from the HEAD response is a good approach to solving this problem. Looking at the implementation, there are a few minor things to improve. Tests are also still missing, which we need to add before this can be merged. Let me know if you need help with this.

Vija02:
Thanks for the review, I've updated them based on your comments. I'll add some tests soon

lib/upload.js
```diff
@@ -821,9 +828,10 @@ class BaseUpload {
       // If the upload length is deferred, the upload size was not specified during
       // upload creation. So, if the file reader is done reading, we know the total
       // upload size and can tell the tus server.
-      if (this.options.uploadLengthDeferred && done) {
+      if (this._uploadLengthDeferred && (!this.options.uploadLengthDeferred || done)) {
```
Acconut, line 831:
```suggestion
      if (this._uploadLengthDeferred || done) {
```

I think we can just use `this._uploadLengthDeferred` here and don't need to consult the original option anymore.

Acconut (changes requested):
One more comment so far. I am looking forward to the tests. Let me know if you need any more help!

lib/upload.js
Vija02: That will change the behaviour quite a bit. As it is suggested right now, there's some problems:
1. We don't always know the size when `this._uploadLengthDeferred` is true since `options.uploadLengthDeferred` could be true. In this case, we'd send wrong info.
2. `done` will make it always pass the header at the end, which I believe isn't how it should work. It should only send the header on the next known request from what I understand.

If we change it to an AND, it'll kinda work, but only send the length header at the end.

The logic in the commit right now expands to:
- `(this._uploadLengthDeferred && !this.options.uploadLengthDeferred)`
 Send header on next opportunity if the request is of known size
- `(this._uploadLengthDeferred && done)`
 And send at the end if it's a stream

Vija02:
@Acconut I've added 2 tests. And also updated a fix for calculating the total size on that code block. Before the fix, it'd calculate the size of the current request and send that rather than the total size

lib/upload.js
Acconut: You're right, my suggestion is wrong. I used `||` instead of `&&` on accident and my suggestion would be:

```suggestion
      if (this._uploadLengthDeferred && done) {
```

> If we change it to an AND, it'll kinda work, but only send the length header at the end.

Yes, that's right and that's how tus-js-client has handled deferred lengths in the past. It always sent the length header once the file source is closed because then we know the file size for sure. I don't see a problem with keeping this behavior. Or do you?

Vija02: Honestly, I'm quite new to the protocol and library so I think you'd know more. But I was looking at this line over at the protocol [documentation](https://tus.io/protocols/resumable-upload#post): 
image

That led me to believe that it should be the first possible instance. Although, I guess both ways should work anyway?

Vija02: Hey @Acconut, just want to bump this PR, any thoughts on this? Thanks 😄

Issue #738, opened 2024-12-19
Make fingerprinting part of FileSource


Acconut:
Currently, fingerprints are calculated through the function that is passed in through the `fingerprint` option, defaulting to following implementation:

https://github.com/tus/tus-js-client/blob/ca63ba254ea8766438b9d422f6f94284911f1fa5/lib/browser/fileSignature.js#L12-L27

As can be seen in the code, the fingerprint function has to differentiate between the different types of sources that tus-js-client can handle. For example, it needs to handle React Native files differently than `File`/`Blob` instances. In addition, it does not contain proper handling for streaming sources.

Generating a fingerprint should not be handled in one central function but be delegated to the [FileSource instances](https://github.com/tus/tus-js-client/blob/main/docs/api.md#filereader), which handle the specific access to different input types. These instance know best how to compute the most accurate fingerprint.

That being said, we should still provider users with the ability to influence the fingerprint and add just information in it.

major-winter:
Hi @Acconut , is this issue still relevant? If so, I would like to work on it.

Acconut:
Yes, I would still like to experiment with this idea and see if it provides the anticipated benefits. It's great that you want to help with this! The way I'm currently imagine it is that the `FileSource` interface is extended with a `fingerprint(): Promise<null 
| string>` method. It either return the fingerprint as a string or `null` if no fingerprint can be calculated, e.g. for streams. The `Upload` can then use this method to retrieve a fingerprint. We should still allow users of tus-js-client to customize the fingerprint from FileSource, e.g. my appending just data to it. For example, if uploads shouldn't be reused across projects, users could append a project ID to make sure that fingerprints of the same file for different projects don't match.

Regarding the calculating of the fingerprint, I've also been thinking about making them more precise. Currently, fingerprints only consist about metadata from the file, which can be inaccurate. We should consider including a checksum of the first few megabytes to ensure that the fingerprint reflects the content properly. But we can tackle this in a separate PR as well and keep the current fingerprint logic while moving it to FileSource.

What do you think? Does this make sense?

Acconut:
Sounds good! Let me know if you need any additional help!

major-winter:
Thanks! Also, how can I assign this item to myself? Just a heads-up—I’ll be on a business trip next week, so there might be a slight delay in delivering the feature

Acconut:
Sure, I can assign you for this. And thanks for the heads-up 👍

PR #774, opened 2025-05-20
Add support for fingerprint generation in FileSource classes

major-winter:
Implemented a `fingerprint` method in various `FileSource` implementations to improve upload tracking.

Introduced project-specific metadata in fingerprints and updated the upload flow for better modularity and support across different platforms, including React Native.

### DIFF
--- a/lib/browser/fileSignature.ts
+++ b/lib/browser/fileSignature.ts
@@ -1,13 +1,9 @@
 import type { ReactNativeFile, UploadInput, UploadOptions } from '../options.js'
-import { isReactNativeFile, isReactNativePlatform } from '../reactnative/isReactNative.js'
 
 /**
  * Generate a fingerprint for a file which will be used the store the endpoint
  */
 export function fingerprint(file: UploadInput, options: UploadOptions) {
-  if (isReactNativePlatform() && isReactNativeFile(file)) {
-    return Promise.resolve(reactNativeFingerprint(file, options))
-  }
 
   if (file instanceof Blob) {
     return Promise.resolve(
@@ -18,25 +14,4 @@ export function fingerprint(file: UploadInput, options: UploadOptions) {
   }
 
   return Promise.resolve(null)
-}
-
-function reactNativeFingerprint(file: ReactNativeFile, options: UploadOptions): string {
-  const exifHash = file.exif ? hashCode(JSON.stringify(file.exif)) : 'noexif'
-  return ['tus-rn', file.name || 'noname', file.size || 'nosize', exifHash, options.endpoint].join(
-    '/',
-  )
-}
-
-function hashCode(str: string): number {
-  // from https://stackoverflow.com/a/8831937/151666
-  let hash = 0
-  if (str.length === 0) {
-    return hash
-  }
-  for (let i = 0; i < str.length; i++) {
-    const char = str.charCodeAt(i)
-    hash = (hash << 5) - hash + char
-    hash &= hash // Convert to 32bit integer
-  }
-  return hash
 }
--- a/lib/node/sources/NodeStreamFileSource.ts
+++ b/lib/node/sources/NodeStreamFileSource.ts
@@ -1,5 +1,5 @@
-import type { Readable } from 'node:stream'
-import type { FileSource } from '../../options.js'
+import { Readable, Transform } from 'node:stream'
+import type { FileSource, UploadOptions } from '../../options.js'
 
 /**
  * readChunk reads a chunk with the given size from the given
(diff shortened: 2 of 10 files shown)


major-winter:
Hi @Acconut,
This is a draft PR—could you please take a look and let me know if I'm heading in the right direction? Thanks!

lib/browser/fileSignature.ts
```diff
@@ -1,13 +1,9 @@
 import type { ReactNativeFile, UploadInput, UploadOptions } from '../options.js'
```
Acconut, line 1: This entire file can probably be removed now, can't it?

lib/node/sources/NodeStreamFileSource.ts
```diff
@@ -74,9 +74,16 @@ export class NodeStreamFileSource implements FileSource {
     })
   }
 
+  // TODO: Implement fingerprint calculation for NodeStreamFileSource.
+  // This should likely involve reading the stream and creating a hash/checksum
+  // while preserving the stream's content for later use.
+  fingerprint(options: UploadOptions): Promise<string | null> {
+    throw new Error("fingerprint not implemented for NodeStreamFileSource")
```
Acconut, line 81: For now, it's best to return `null` to keep the existing behavior and align with `WebStreamFileSource`.

lib/node/sources/PathFileSource.ts
```diff
@@ -34,6 +34,14 @@ export class PathFileSource implements FileSource {
     this.size = size
   }
 
+  fingerprint(options: UploadOptions): Promise<string | null> {
+    if (typeof options.fingerprint === 'function') {
+      return Promise.resolve(options.fingerprint(this._file, options))
+    }
```
Acconut, line 40: Allowing users to modify the fingerprint is an important aspect to ensure that tus-js-client is customizable. My vision for the current `fingerprint` option is that its a signature like this:

```
fingerprint: (file: UploadInput, options: UploadOptions, sourceFingerprint: string | null) => Promise<string | null>
```

The callback gets the fingerprint from the file source passed and can then decide to modify it (e.g. by appending information) or to overwrite it completely.

Invoking this callback is best handled in the Upload class, so it doesn't have to be done in every implementation of FileSource. Does this make sense?

lib/options.ts
```diff
@@ -51,6 +51,8 @@ export type UploadInput =
 export interface UploadOptions {
   endpoint?: string
 
+  projectId?: string
```
Acconut, line 54: Can you explain what this option is about?

lib/sources/WebStreamFileSource.ts
```diff
@@ -39,6 +39,8 @@ export class WebStreamFileSource implements FileSource {
 
   private _done = false
 
+  private _stream: ReadableStream
```
Acconut, line 42: This property doesn't seem used. Can it be removed?

Acconut (changes requested):
Thank you very much for getting started on this! I think it's going in the right direction and I left a few comments to hopefully help you progress.

lib/options.ts
major-winter: To allow users to differentiate the same file across multiple projects, they could append a project ID to the upload so that fingerprints of the same file in different projects don't match. As you described it [here](https://github.com/tus/tus-js-client/issues/738#issuecomment-2841200871).

Does this align with your idea, or do you envision a different approach?

lib/node/sources/PathFileSource.ts
major-winter: Let me summarize how I understand the direction:

- We’ll extend the fingerprint option signature to include a `sourceFingerprint` parameter.

- The fingerprint callback can then decide whether to use the `sourceFingerprint` as-is or modify it (e.g., by appending a project ID).

- This callback will be implemented centrally in the `Upload` class for each environment (Node and browser), so that individual `FileSource` classes won’t need to handle fingerprint logic themselves.

lib/browser/fileSignature.ts
major-winter: This file has been removed.

lib/node/sources/NodeStreamFileSource.ts
major-winter: I've modified the method to return a promise that resolves to null.

lib/options.ts
major-winter: This field has been removed.

lib/sources/WebStreamFileSource.ts
major-winter: this property has been removed.

Acconut:
Thanks for the updates. I'll have a look at them in 1-2 weeks.

PR #775, opened 2025-05-20
feat: Add stall detection to recover from frozen uploads

m7kvqbe1:
## Overview 

This feature addresses the issue of uploads hanging indefinitely in unreliable network conditions, particularly in Node.js environments where no default timeout exists.

When uploads stall due to network issues, TCP connections can enter a degraded state where no data is transferred but no error is triggered. This implementation detects such stalls and forces a retry.

## Implementation details

- Supports two detection methods:
  - Progress-based: Detects when no upload progress events are fired
  - Rate-based: Detects when overall transfer rate drops below threshold
- Automatically selects the appropriate method based on HTTP stack capabilities
- Gracefully integrates with the existing retry mechanism
- Fully configurable with sensible defaults:
  - 30s stall timeout (time with no progress before considering stalled)
  - 5s check interval (how often to check for stalls)
  - 1 byte/s minimum transfer rate

This is especially important for uploads over satellite links, cellular networks, or other unreliable connections where TCP backoff can cause indefinite stalls.


m7kvqbe1:
More tests coming soon.

lib/upload.ts
```diff
@@ -985,6 +1014,133 @@ export class BaseUpload {
...
+  }
+
+  /**
+   * Detect if current HttpStack supports progress events
+   */
+  private _supportsProgressEvents(): boolean {
+    const httpStack = this.options.httpStack
+    // Check if getName method exists and if it returns one of our known stacks
+    return typeof httpStack.getName === 'function' &&
+           ["NodeHttpStack", "XHRHttpStack"].includes(httpStack.getName())
```
Acconut, line 1039: Instead of checking the name of the HTTP stack, it's better if the stack exposes its progress reporting capabilities with a newly added property or method.

lib/upload.ts
```diff
@@ -985,6 +1014,133 @@ export class BaseUpload {
...
+  }
+
+  /**
+   * Check if upload has stalled based on progress events
+   */
+  private _isProgressStalled(now: number): boolean {
```
Acconut, line 1045: The Upload class is already pretty large with many properties and methods. Stall detection appears to be a part that can be moved into its own component, making testing easier and growing `upload.ts` by fewer lines. Could you move the logic into a seperate component and file?

lib/upload.ts
```diff
@@ -985,6 +1014,133 @@ export class BaseUpload {
...
+   * Apply default stall detection options
+   */
+  private _getStallDetectionDefaults(
+    options?: Partial<StallDetectionOptions>
+  ): StallDetectionOptions {
+    return {
+      enabled: options?.enabled ?? true,
+      stallTimeout: options?.stallTimeout ?? 30000,
+      checkInterval: options?.checkInterval ?? 5000,
+      minimumBytesPerSecond: options?.minimumBytesPerSecond ?? 1
```
Acconut, line 1028: Default option values should be set in `defaultOptions`, so they are all in one place and also properly exposed to users. That being said, I'm not sure if we want to enable this by default in the next release already as we still have to gain experience with this approach and tune it properly. False positive (incorrectly reported stalls) can cause major disruption for tus-js-client users.

Acconut (commented):
Thank you for working on this, Tom! It's greatly appreciated! I'm excited about this feature, and we have been kicking around this idea for some time already.

I had a high level look at your PR and left a few comments. 

> Rate-based: Detects when overall transfer rate drops below threshold

This mechanism will likely only work if the upload is split across multiple requests, won't it? If a large file is uploaded at normal speeds over 5min in one request, but the HTTP stack does not support progress events, it will be detected as a stall. Such a scenario is not distinguishable from actual stalls.

If progress events are not available and one wants to have stall detection, the best option would be to enable `chunkSize` and set an appropriate request timeout, which fits the chunk size.

In addition, requests can be delayed for reasons other than stalls. For example, some people use the `onBeforeRequest` callback to implement a rate limit/queue system for outgoing requests. Or they fetch authentication tokens to be uses in the upload requests. These processes might delay requests and thus progress events, so we have to be careful to not flag them as stalls.

Let me know what you think!

lib/upload.ts
m7kvqbe1: Would you consider enabling it via an experimental flag or something - or maybe doing a pre-release?

m7kvqbe1: We have some good tooling for testing it, we have created a containerised satellite network emulator proxy.

Once the PR starts to solidify I will be putting it through its paces with that through varying network conditions.

I've got a test matrix I can run over it as part of a nightly job.

Acconut: > Would you consider enabling it via an experimental flag or something

Not sure I understand. I'm looking forward to landing this feature in the next releases, but I would just not enable it default for now. If user want to utilize the stall detection, they can always enable this feature using the options.

> Once the PR starts to solidify I will be putting it through its paces with that through varying network conditions.
> I wonder if you'd also be interested in using this tooling?
> I'll open a discussion internally about open sourcing it today.

Absolutely! That sounds fantastic!

lib/options.ts
```diff
@@ -141,6 +158,9 @@ export type SliceResult =
 export interface HttpStack {
   createRequest(method: string, url: string): HttpRequest
   getName(): string
+  // Indicates whether this HTTP stack implementation supports progress events
+  // during upload. If false, stall detection will use overall transfer rate instead.
+  supportsProgressEvents?: () => boolean
```
Acconut, line 163: This function doesn't have to be optional as this PR would be released in a new major release, where we can make breaking changes.

lib/options.ts
```diff
@@ -48,42 +48,59 @@ export type UploadInput =
...
+export type Part = { start: number; end: number }
 
-  onProgress?: (bytesSent: number, bytesTotal: number | null) => void
-  onChunkComplete?: (chunkSize: number, bytesAccepted: number, bytesTotal: number | null) => void
-  onSuccess?: (payload: OnSuccessPayload) => void
-  onError?: (error: Error | DetailedError) => void
-  onShouldRetry?: (error: DetailedError, retryAttempt: number, options: UploadOptions) => boolean
-  onUploadUrlAvailable?: () => void | Promise<void>
+export interface UploadOptions {
+  endpoint?: string | null
```
Acconut, line 63: A lot of the changes in this file seem unrelated to the PR, such as these type changes. Can you please remove them (or propose them in a separate PR) to keep this one focused?

lib/upload.ts
```diff
@@ -27,17 +33,17 @@ export const defaultOptions = {
   fingerprint: undefined,
   uploadSize: undefined,
 
-  onProgress: undefined,
-  onChunkComplete: undefined,
-  onSuccess: undefined,
-  onError: undefined,
-  onUploadUrlAvailable: undefined,
+  onProgress: null,
```
Acconut, line 36: Similarly to the comment above, these changes seem unrelated.

lib/upload.ts
```diff
@@ -215,6 +230,9 @@ export class BaseUpload {
       }
     }
 
+    // Setup stall detection before starting the upload
+    this._setupStallDetection()
```
Acconut, line 234: This PR currently uses one stall detector for the entire upload, which seems to increase the complexity of the stall detection as it now has to support pausing and resuming as well.

A simpler alternative would be to use to stall detector per HTTP request that carries upload data (i.e. PATCH or POST requests with `uploadDataDuringCreation`). Then the stall detector has to keep less state as we just discard it when the request finishes or is aborted, but we should still be capable of detecting stalled uploads.

Acconut (changes requested):
Thanks for the updates. I had another look and left some comments. Overall, this is moving in the right direction!

lib/upload.ts
m7kvqbe1: @Acconut 

This should now have been addressed along with the other feedback. Feel free to take another look.

❤️

test/spec/node-index.js
```diff
@@ -5,3 +5,4 @@ import './test-terminate.js'
 import './test-web-stream.js'
 import './test-binary-data.js'
 import './test-end-to-end.js'
+import './test-stall-detection.js'
```
Acconut, line 8: Is this test specific to Node.js? If not, it should be added to `test-common.js`, so it's also run in automated browsers.

test/spec/test-stall-detection.js
```diff
@@ -0,0 +1,425 @@
...
+// Helper to get body size
+function getBodySize(body) {
+  if (body == null) return null
+  if (body instanceof Blob) return body.size
+  if (body.length != null) return body.length
+  return 0
+}
+
+describe('tus-stall-detection', () => {
+  describe('configuration options', () => {
```
Acconut, line 45: I greatly appreciate you writing these configuration tests, but I am wondering if they are truly necessary here. They mostly seem to check that the option merging is working properly. Bug in there would also be detected in the integration tests, so I think we can use some LOC here :)

test/spec/test-stall-detection.js
```diff
@@ -0,0 +1,425 @@
...
+  constructor() {
+    super()
+    this.stallOnNextPatch = false
+  }
+
+  createRequest(method, url) {
+    const req = super.createRequest(method, url)
+
+    // If we want to simulate a stall on PATCH requests, override the send method
+    if (this.stallOnNextPatch && method === 'PATCH') {
```
Acconut, line 15: I think it would be useful to have a test that not only checks if requests without any progress events are detected as well, but also requests where the progress events stop in the middle of the request.

lib/StallDetector.ts
```diff
@@ -0,0 +1,112 @@
...
+    )
+  }
+
+  /**
+   * Check if upload has stalled based on progress events
+   */
+  private _isProgressStalled(now: number): boolean {
+    const timeSinceProgress = now - this.lastProgressTime
+    const stallTimeout = this.options.stallTimeout
+    const isStalled = timeSinceProgress > stallTimeout
```
Acconut, line 95: I wonder if we should not only check the elapsed time since the last progress, but also the progress value itself. For example, if a HttpStack regularly reports the some progress of X bytes uploaded, the stall detector will currently not catch that.

Acconut (changes requested):
Thank you for the updates! I left a few more comments to make this detection extra robust!

m7kvqbe1:
@Acconut 

Feel free to review again, all that feedback should now have been addressed.

We'd really love to get our hands on this as soon as possible. What do you think about shipping a pre-release? 

We have a system we can test this with now and a load of established test automation that leverages a containerised sat-link emulation proxy (we run a matrix of dynamic adverse network scenarios and transfers of varying sizes).

I hope to open-source that tooling at some point so you guys can use it too.

m7kvqbe1:
@Acconut those bits should now have been addressed, feel free to take another look.

Acconut:
> Do you think we might be able to ship a pre-release of this before the end of the week?

I'll try, but I cannot make any promises at the time.

Acconut:
> No rush from our side, i've published a release of the fork to test with a real-world consuming service.

That's great and is probably the best way to take this for test drive right now.

> Basically in my real-world tests this morning i've discovered we can't get low level TCP socket transfer progress from the Node HTTP module, you can only see an internal I/O buffer being filled, you then get silence until that buffer is completely drained (transmitted) and starts to be filled again, this means you consistently get false positives for stall detection.

Oh, I see. Getting progress information in Node.js has been rather tedious compared to XMLHttpRequest in browsers, which provides progress events on its own. In Node.js, we use our own logic to write smaller chunks to the HTTP request stream to get progress information, see https://github.com/tus/tus-js-client/blob/021960d557ec24db283041eb425ff12902d73305/lib/node/NodeHttpStack.ts#L225-L276.

But as you correctly mentioned, this only allows us to retrieve details about how much data landed in the HTTP request stream. There are usually other buffers involved before the data hit the networks. We can't even assume that the HTTP request has its own TCP connection since HTTP/2 and HTTP/3 perform multiplexing over TCP and UDP, respectively. So even if there was a way to get low-level TCP information, it would only be useful for HTTP/1.1 requests.

That being said, I wonder if you could prevent false positives by tuning the stall detection parameters. Increasing the stall timeout while reducing the high water mark in the HTTP request stream (not sure how that's best done, to be honest) could help. In addition, you might want to experiment with different chunk sizes or throttle configurations in `writeBufferToStreamWithProgress` to get the desired behavior.

In addition, I tested the stall detection on my own yesterday and it does not seem to fully work in browsers yet. The logic assumes that our HTTP stacks emit an error when requests are aborted, but the XHRHttpStack does not do that. It just leaves the promise hanging. This should be easy to fix and I'm working on a patch, but more testing is needed.

I also learned that [Uppy](https://uppy.io) has a similar approach to stall detection: https://github.com/transloadit/uppy/blob/main/packages/%40uppy/utils/src/ProgressTimeout.ts It's a bit simpler as it doesn't use check intervals, but just always renew the timeout when a progress event is emitted. It's also enabled by default with a timeout of 30s and they see it's working well. Might be worth a look, although Uppy is only designed for browsers and not Node.js.

Acconut:
> In addition, I tested the stall detection on my own yesterday and it does not seem to fully work in browsers yet. The logic assumes that our HTTP stacks emit an error when requests are aborted, but the XHRHttpStack does not do that. It just leaves the promise hanging. This should be easy to fix and I'm working on a patch, but more testing is needed.

Opened https://github.com/tus/tus-js-client/pull/778 for this.

PR #778, opened 2025-06-13
Emit errors when request is aborted

Acconut:
While working on https://github.com/tus/tus-js-client/pull/775, we discovered that not every HTTP stack in tus-js-client behaves the same way when aborting requests:

- XHRHttpStack just leaves the promise returned by `send` hanging, neither resolving nor rejecting it.
- NodeHttpStack threw a network-related error (usually ECONNRESET).
- FetchHttpStack threw a DOMException.

While the exact error is not important, it's helpful to standardize on throwing some error, as allows for simpler logic when aborting requests. Hence this PR, which implements this. It uses a DOMException since normal Error instances in JavaScript don't have proper error codes. While DOMException is a Web API, I think it's available in most JavaScript runtimes.


m7kvqbe1:
LGTM

Issue #787, opened 2025-07-02
Have a "shouldRemoveFromUrlStorage" function


reinforceColumn:
**Is your feature request related to a problem? Please describe.**
During the start of an upload the client does a HEAD request to know if file already exists on server and know how much of it is already uploaded. If this request fails (for example for an authentication error), it's not possible to stop the file from being deleted.

**Describe the solution you'd like**
Have a way to handle the errors thrown by the HEAD request  

**Describe alternatives you've considered**
Another possibility would be to decide before a file is deleted if it actually needs to be deleted

Acconut:
Good point, the client should provide users with more controls in such a situation. Would you be open to work on a PR for this?

reinforceColumn:
I cannot say i can find time for it but i'll try

Acconut:
I can understand that. Let me know if you need help at any stage.

PR #788, opened 2025-07-04
Added onShouldRemoveFromUrlStorage functionality

reinforceColumn:
draft for #787


reinforceColumn:
It would be useful to add req/res or DetailedError, but it's called in multiple instances and don't know the general policy

PR #812, opened 2025-11-14
WIP: Make Header Indexing Case-Insensitive in tus Client

daltonpearson:
I realize this may be somewhat contentious, but I wanted to start with a workable solution before opening the discussion.

According to the HTTP/1.1 specification, header field names are case-insensitive, while in HTTP/2 they are required to be lowercase.

Currently, the tus client relies on case-sensitive indexing of headers, which causes issues in environments where backend stacks normalize headers to lowercase — a behavior already adopted by many frameworks and servers.

This PR introduces support for case-insensitive header lookups by normalizing header keys to lowercase when performing lookups.

---

⚙️ Summary

- Ensures compatibility with HTTP/1.1 and HTTP/2 expectations

- Fixes issues in backend setups where headers are automatically rewritten to lowercase

- Aligns tus client behavior with common practice across HTTP client and server libraries

---

💬 Feedback Welcome

If there’s another approach the maintainers would prefer, please let me know — I’m happy to adjust the implementation to align with your preferred style or design direction.

### DIFF
--- a/lib/browser/FetchHttpStack.ts
+++ b/lib/browser/FetchHttpStack.ts
@@ -93,7 +93,7 @@ class FetchResponse implements HttpResponse {
   }
 
   getHeader(header: string): string | undefined {
-    return this._res.headers.get(header) || undefined
+    return this._res.headers.get(header) || this._res.headers.get(header.toLowerCase()) || undefined
   }
 
   getBody(): string {
--- a/lib/browser/XHRHttpStack.ts
+++ b/lib/browser/XHRHttpStack.ts
@@ -112,7 +112,11 @@ class XHRResponse implements HttpResponse {
   }
 
   getHeader(header: string): string | undefined {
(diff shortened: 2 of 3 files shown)


daltonpearson:
@Acconut Our backend stack is normalizing the headers eg Location -> location, the problem is that tus is looking for the header "Location" instead of the normalized "location" therefore tus throws an error saying that the location/Location header is not set and fails.

Acconut:
That's odd because by default tus-js-client uses XMLHttpRequest in browsers, where the search for headers is case-insensitive: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getResponseHeader

The same is true for the Fetch API: https://developer.mozilla.org/en-US/docs/Web/API/Headers/get

I'm still not able to understand where exactly this problem seems to arise.

PR #859, opened 2026-04-02
Remove redundant url-parse dependency in favor of native URL API

RinZ27:
The codebase included an import for `url-parse` that was effectively dead code, as the core logic was already utilizing the native WHATWG URL API. After reviewing the implementation, I noticed a TODO mentioning the desire to replace this CommonJS package to streamline the ESM transition and remove the need for specific Rollup plugins.

This change removes `url-parse` from `dependencies` and updates the `resolveUrl` utility to handle relative origins gracefully using the native `URL` API. This is particularly important for test environments where endpoints might be defined as relative paths. Removing this dependency not only eliminates several reported security vulnerabilities (CVE-2022-0686, CVE-2022-0691) but also reduces the bundle size and simplifies the build pipeline.

I've verified the changes against the existing test suite, and the logic remains compatible with both absolute and relative URL resolutions.

### DIFF
--- a/lib/upload.ts
+++ b/lib/upload.ts
@@ -1,7 +1,4 @@
 import { Base64 } from 'js-base64'
-// TODO: Package url-parse is CommonJS. Can we replace this with a ESM package that
-// provides WHATWG URL? Then we can get rid of @rollup/plugin-commonjs.
-import URL from 'url-parse'
 import { DetailedError } from './DetailedError.js'
 import { log } from './logger.js'
 import {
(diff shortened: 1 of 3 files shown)


Acconut:
Thank you for the PR, but we cannot get rid of the dependency without breaking support for React Native, where `URL` isn't available by default. Please see https://github.com/tus/tus-js-client/pull/681 for more details.

RinZ27:
@Acconut, thanks for the clarification and for pointing me to #681. I admittedly overlooked the React Native environment's lack of a native WHATWG URL implementation. Given that maintaining full platform compatibility is a priority for tus-js-client, it's clear that this refactor is counterproductive. I'll go ahead and close this PR.

PR #860, opened 2026-04-07
Wrap errors thrown in user callbacks with descriptive messages

### LINKED ISSUE #294: Improve error message when error occurs in user's callback
Consider following code:

```js
input.addEventListener("change", function(e) {
    // Get the selected file from the input element
    var file = e.target.files[0]

    // Create a new tus upload
    var upload = new tus.Upload(file, {
        endpoint: "http://localhost:1080/files/",
        retryDelays: [0, 3000, 5000, 10000, 20000],
        metadata: {
            filename: file.name,
            filetype: file.type
        },
        onError: function(error) {
            console.log("Failed because: " + error)
        },
        onProgress: function(bytesUploaded, bytesTotal) {
            // This will cause an error
           someThing.someProperty
        },
        onSuccess: function() {
        }
    })

    // Start the upload
    upload.start()
})
```

This will cause the upload to fail with an error, such as `Error: tus: failed to upload chunk at offset 0, caused by TypeError: Cannot read properties of undefined (reading 'someProperty'), originated from request`. It looks like it is an internal failure of tus-js-client but actually it is coming from the user's callback, confusing the end user. This happened in #293 and #330, for example.

I propose to catch error when callbacks are invoked and bubble them up with a proper error description explaining the cause.

Working callbacks:
- [x] onSuccess
- [ ] onError
- [ ] onProgress
- [ ] onChunkComplete
- [ ] onShouldRetry
- [ ] onBeforeRequest
- [ ] onAfterResponse

vikas90244:
Fixes #294

When a user's callback (onProgress,  onError, onChunkComplete, onBeforeRequest, onAfterResponse, onShouldRetry) throws an error, the error  message now clearly indicates which callback caused it instead of appearing  as an internal tus error.

### DIFF
--- a/lib/upload.ts
+++ b/lib/upload.ts
@@ -494,7 +494,12 @@ export class BaseUpload {
     if (this._aborted) return
 
     if (typeof this.options.onError === 'function') {
-      this.options.onError(err)
+      try {
+        this.options.onError(err)
+      } catch (callbackErr) {
+        const causeMsg = callbackErr instanceof Error ? callbackErr.message : String(callbackErr)
+        throw new Error(`tus: error thrown in 'onError' callback: ${causeMsg}`)
+      }
     } else {
       throw err
     }
@@ -558,7 +563,12 @@ export class BaseUpload {
    */
   private _emitProgress(bytesSent: number, bytesTotal: number | null): void {
     if (typeof this.options.onProgress === 'function') {
-      this.options.onProgress(bytesSent, bytesTotal)
+      try {
+        this.options.onProgress(bytesSent, bytesTotal)
+      } catch (err) {
+        const causeMsg = err instanceof Error ? err.message : String(err)
+        throw new Error(`tus: error thrown in 'onProgress' callback: ${causeMsg}`)
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


lib/upload.ts
```diff
@@ -494,7 +494,12 @@ export class BaseUpload {
     if (this._aborted) return
 
     if (typeof this.options.onError === 'function') {
-      this.options.onError(err)
+      try {
+        this.options.onError(err)
+      } catch (callbackErr) {
+        const causeMsg = callbackErr instanceof Error ? callbackErr.message : String(callbackErr)
+        throw new Error(`tus: error thrown in 'onError' callback: ${causeMsg}`)
```
Acconut, line 501: I believe we can set the cause property to expose the original error to the caller.

```suggestion
        throw new Error(`tus: error thrown in 'onError' callback: ${causeMsg}`, { cause: callbackErr })
```

Acconut, line 501: Is it possible to add a test case for this? So we know that errors from callbacks are properly caught, wrapped and don't send the upload logic into a retry loop.

Acconut (commented):
Thank you very much for working on this improvement! I have two minor comments, but otherwise this looks good :)

lib/upload.ts
vikas90244: Thanks for the suggestions!

I tried using `{ cause: callbackErr }` in the `Error` constructor but it results in `"Expected 0-1 arguments, but got 2"` since ErrorOptions is not available with the current ES2018 target.

As an alternative I considered manually assigning `wrappedErr.cause = callbackErr` but TypeScript also rejects this without updating the lib compiler option.
For now I've kept the implementation without .cause,  the original error message is already included in the wrapping message string via causeMsg, so the information is still visible to the caller. Happy to revisit if you have a preferred approach.

vikas90244: working on it.

vikas90244:
Hi  @Acconut  , just a gentle bump on this PR when you get a chance. Thanks!

PR #866, opened 2026-04-12
fix: upload stalls with deferred length when size is multiple of chunkSize

### LINKED ISSUE #779: Upload stalls when using deferred length and stream size is exactly divisible by chunkSize
**Describe the bug**
When using Readable streams with uploadLengthDeferred: true, the upload hangs if the total stream size is an exact multiple of chunkSize. This issue does not occur when:

The total stream size is not divisible by chunk size.

The upload uses a known Upload-Length.

**To Reproduce**
This code snippet from the client side would reproduce the issue
`const input = new stream.PassThrough();
input.end("my hello WORLD"); // 14 bytes

const upload = new Upload(input, {
  endpoint: "/uploads",
  chunkSize: 7,
  uploadLengthDeferred: true,
  onSuccess: () => console.log("Upload complete"),
});
upload.start();`

**Expected behavior**
Upload completes with Upload-Length sent in the final PATCH.

**Actual behavior** 
Upload hangs after final chunk if size % chunkSize === 0.

**Setup details**

- Runtime environment: Node.js 
- Used tus-js-client version: 4.3.1
- Used tus server software: tus/server

Yanhu007:
Fixes #779

## Problem

When using `uploadLengthDeferred: true` with a Node.js stream whose total size is exactly divisible by `chunkSize`, the upload hangs forever after the last chunk.

**Root cause:** After consuming the last aligned chunk, `readChunk()` is called again. The stream has no more data and emits `end`, but `readChunk` only listens for `readable` events — the promise never resolves.

### Timeline

1. `slice(0, 7)` → reads 7 bytes, `done: false`
2. `slice(7, 14)` → reads 7 bytes, stream buffer depleted, `done: false`
3. `slice(14, 21)` → buffer empty, `readChunk()` called → **hangs** (stream emits `end`, not `readable`)

## Fix

- Add `end` event listener in `readChunk()` that resolves with an empty buffer
- Add `readableEnded` guard at entry for already-ended streams
- Clean up all three listeners (`error`, `readable`, `end`) on resolution

### DIFF
--- a/lib/node/sources/NodeStreamFileSource.ts
+++ b/lib/node/sources/NodeStreamFileSource.ts
@@ -12,9 +12,22 @@ import type { FileSource } from '../../options.js'
  */
 function readChunk(stream: Readable, size: number) {
   return new Promise<Buffer>((resolve, reject) => {
+    // If the stream has already ended, resolve immediately with an empty buffer.
+    if (stream.readableEnded) {
+      resolve(Buffer.alloc(0))
+      return
+    }
+
     const onError = (err: Error) => {
       stream.off('readable', onReadable)
+      stream.off('end', onEnd)
       reject(err)
+    }
+
+    const onEnd = () => {
+      stream.off('error', onError)
+      stream.off('readable', onReadable)
+      resolve(Buffer.alloc(0))
     }
 
     const onReadable = () => {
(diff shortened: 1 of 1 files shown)


Acconut:
Thank you for this report and providing a fix as well. Could you add a unit test covering this case so we'll catch any future regression?

PR #881, opened 2026-07-13
docs: explain proxy body-size limits and how to pick a chunkSize

junkerderprovinz:
The `chunkSize` documentation already lists "a tus server or proxy with a limit on how big request bodies may be" as one of the two valid reasons for setting the option, but it does not describe what actually happens when you hit such a limit, nor which value to pick.

We ran into this in production: uploads behind a proxy failed with a 413 status code for larger files, and nothing pointed towards `chunkSize` as the cause. With the default of `Infinity` the whole file goes into a single `PATCH` request, and the proxy rejects it before the tus server ever sees it. Raising the origin's own body limits does not help, because the proxy caps the request first.

This PR expands that one bullet with the mechanism, the symptom (a 413, or a closed connection), and the guidance to pick a chunk size below the smallest body-size limit on the path to the server, while respecting server-side minimums (the S3 note that already follows).

Following the review, the vendor names and their current limits are gone. Only the concept remains.

No code changes.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -143,7 +143,7 @@ A number indicating the maximum size of a `PATCH` request body in bytes. The def
 **Warning:** **Do not set this value**, unless you are being forced to. The only two valid reasons for setting `chunkSize` are:
 
 - You are passing a reader or readable stream as input to tus-js-client and it will complain that it "cannot create source for stream without a finite value for the chunkSize option" if you leave `chunkSize` empty.
-- You are using a tus server or proxy with a limit on how big request bodies may be.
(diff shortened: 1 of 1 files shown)


junkerderprovinz:
Ping on this docs-only change. It expands the existing chunkSize bullet in docs/api.md with what happens behind a proxy body-size limit and how to pick a value, which is what cost us time when Cloudflare's 100 MB cap was rejecting uploads with a 413. One line in one file and all checks pass; happy to cut the wording down, or close it if you'd rather keep that bullet terse.

Acconut (changes requested):
Thank you for helping us improve the documentation, but it's not a good idea to place limits of products from unrelated companies in our documentation. These values will change over time and their our docs are out-of-date. It's better to explain the concept (i.e. that some proxies limit request bodies), so that users can figure out if that affects them or not.

junkerderprovinz:
Good point, thanks. I dropped the vendor names and the numbers and kept only the mechanism: a proxy may cap the request body, the single `PATCH` from the default `chunkSize` runs into that cap, and the tus server never sees the request. Branch updated.

Acconut (approved):

Acconut:
Thank you very much for this helpful contribution!

PR #899, opened 2026-09-03
Make tus-js-client usable in a Service Worker

gap777:
Motivation

We upload proof-of-placement photos from a PWA used by contractors on weak or absent mobile networks, iOS Safari being the primary target. The page captures a photo into IndexedDB and a Service Worker drains the queue, so an upload survives the page being closed and can be woken by Background Sync.

That architecture is what pushed us to tus: a Service Worker has no XMLHttpRequest, so upload.onprogress does not exist there, and a connection dropped mid-upload restarts from zero. Chunked, resumable uploads answer both — with a finite chunkSize, chunk completion is the progress signal, so the absence of upload-progress events stops mattering.

We could not use tus-js-client in that context. This PR is the set of changes that made it work.

The constraint: the library cannot currently load in a Service Worker

Not "does not report progress" — does not load at all.

import { Upload } from 'tus-js-client'
// ReferenceError: window is not defined

lib/browser/urlStorage.ts runs its feature detection at module scope:

hasStorage = 'localStorage' in window

A Service Worker (and a Web Worker) has no window, so this throws a ReferenceError. The surrounding catch only converts DOMException — the SecurityError from a sandboxed iframe and the QuotaExceededError from iOS private mode — and rethrows everything else. So the ReferenceError escapes, and because lib/browser/index.ts imports this module, the throw takes down the entire browser entry point on import.

The comment above the line explains the intent:

// Note: localStorage does not exist in the Web Worker's context, so we must use window here.

The premise is right and the conclusion is inverted: window does not exist in a worker either, which is exactly what makes the line throw. This affects both the 4.x line and 5.0.0-pre2.

Reproduced with a bundle evaluated with self defined and window absent, which is what a Service Worker global scope looks like:

before: ReferenceError: window is not defined
after:  loads

What is in this PR

1. Fix urlStorage throwing on import outside a document

typeof localStorage !== 'undefined' instead of 'localStorage' in window. It is false where there is no storage, and it still surfaces the SecurityError that a sandboxed iframe raises from the property getter, so the private-mode and iframe handling below is unchanged. The storage probe becomes conditional, since it cannot run where there is no storage.

Regression test lives in the Node suite, which is the one environment in the project where window is genuinely absent. It imports lib.esm/browser/urlStorage.js directly and asserts canStoreURLs === false.

▎ Importing the whole browser entry point in Node would be the better test, but it hits an unrelated wall first: is-stream@2 is CJS and Node's native ESM cannot take a named readable import from it. Bundlers handle that interop, so it only affects a raw Node import. Happy to widen the test if you would rather solve that too.

2. Add tests for FetchHttpStack

FetchHttpStack shipped with // TODO: Add tests for this. and no coverage. Nine specs covering the HttpStack / HttpRequest / HttpResponse contract: name, method and URL accessors, header round-tripping, what reaches fetch(), the status, headers and body it exposes back, the absent underlying request object, the documented no-op progress handler, and abort() rejecting the send.

The stub rejects with signal.reason rather than a fabricated AbortError, matching what the real Fetch API does — which is what lets the timeout test below distinguish the two.

3. Add an optional timeout to FetchHttpStack

The Fetch API has no timeout of its own, so a request that stalls without erroring is never resolved nor rejected. A custom XHRHttpStack caller can reach for XMLHttpRequest.timeout; a FetchHttpStack caller had nothing, and there is no timeout in UploadOptions either.

This matters most in the environment FetchHttpStack exists to serve. In a Service Worker there is no page to notice a hung PATCH, so the upload simply strands.

new FetchHttpStack({ timeout: 30000 })

It aborts with a TimeoutError DOMException rather than the default AbortError, so a caller can tell a deadline from an abort() it asked for and decide whether the attempt is worth retrying. Defaults to no timeout, so existing behaviour is unchanged.

Also documents FetchHttpStack in docs/api.md, which covered the HttpStack interface but never the fetch implementation of it — including the two things that surprise people: setProgressHandler is a no-op, so chunkSize is what determines progress granularity, and Node readable streams are not supported as a body.

Testing

- Browser (karma/puppeteer): 81 specs, green
- Node (jasmine): 80 specs, green

One note in case it saves someone time: the Chromium that puppeteer.executablePath() resolves to was a stale x86 build on an arm64 machine, and karma fails with Cannot start ChromeHeadless. Setting CHROME_BIN to a system Chrome works.

Open questions for maintainers

1. isSupported still requires XMLHttpRequest. With FetchHttpStack the library now works where XMLHttpRequest does not exist, so isSupported reports false in a Service Worker despite the library being usable there. I left the semantics alone rather than change them unilaterally. Would you like it widened, or a separate export?

2. Default httpStack in worker contexts. lib/browser/index.ts still defaults to XHRHttpStack, so worker callers must pass httpStack explicitly. That seems right to me — it keeps the default path unchanged for every existing consumer — but an environment-sensitive default is an option if you would prefer one.

3. Should the timeout live on UploadOptions instead? A stack-level option was the smaller change, but a timeout that every stack honours may be the better shape. Happy to move it.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -370,6 +370,34 @@ interface HttpResponse {
 
 ```
 
+##### FetchHttpStack
+
+Browsers ship two implementations. The default, `XHRHttpStack`, is built on `XMLHttpRequest` and reports upload progress. `FetchHttpStack` is built on the Fetch API and works where `XMLHttpRequest` does not exist — most notably inside a service worker:
+
+```js
+import { Upload } from 'tus-js-client'
+import { FetchHttpStack } from 'tus-js-client/browser/FetchHttpStack'
+
+const upload = new Upload(file, {
+  endpoint: '/files',
+  httpStack: new FetchHttpStack(),
+  chunkSize: 512 * 1024,
+})
+```
+
+Two differences from `XHRHttpStack` are worth knowing:
+
+- `setProgressHandler` is a no-op, because the Fetch API exposes no upload progress. `onProgress` still fires once per completed `PATCH` request, so a finite `chunkSize` is what determines how granular progress is.
+- Node.js readable streams are not supported as a request body.
+
+`FetchHttpStack` accepts an optional `timeout`, in milliseconds:
+
+```js
+new FetchHttpStack({ timeout: 30000 })
+```
+
+The Fetch API has no timeout of its own, so without one a request that stalls without erroring is never resolved nor rejected. A request cancelled by the timeout rejects with a `TimeoutError` `DOMException`, which is distinguishable from the `AbortError` that `abort()` produces.
+
 #### urlStorage
 
 _Default value:_ Environment-specific implementation
--- a/lib/browser/FetchHttpStack.ts
+++ b/lib/browser/FetchHttpStack.ts
@@ -7,10 +7,27 @@ import type {
   SliceType,
 } from '../options.js'
 
-// TODO: Add tests for this.
+export interface FetchHttpStackOptions {
+  /**
+   * Milliseconds to wait for a request to complete before it is aborted. The
+   * Fetch API has no timeout of its own, so without this a request that stalls
+   * without erroring is never resolved nor rejected. Defaults to no timeout.
+   *
+   * A request cancelled this way rejects with a `TimeoutError` DOMException,
+   * which is distinguishable from the `AbortError` that `abort()` produces.
+   */
+  timeout?: number
+}
+
 export class FetchHttpStack implements HttpStack {
+  private _timeout?: number
+
+  constructor({ timeout }: FetchHttpStackOptions = {}) {
+    this._timeout = timeout
+  }
+
   createRequest(method: string, url: string) {
-    return new FetchRequest(method, url)
+    return new FetchRequest(method, url, this._timeout)
   }
 
   getName() {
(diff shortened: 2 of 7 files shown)


gap777:
@Acconut Can I ask you to review this, and if acceptable, merge it?

Acconut:
Thank you for opening this PR, @gap777. The PR seems to mix unrelated changes together, which makes it harder to review and understand. Unfortunately, the PR description doesn't help me much as it's too convoluted and doesn't clearly express the problem and its solution.

As far as I understand it, the root problem is that there was some reference to `window`, which isn't available in the context of a service worker. This PR should be focused on addressing this and nothing else. The added tests for FetchHttpStack are useful but should go into another PR.

I'm not entirely sure about the `timeout` option yet. It shouldn't be an option for one specific HTTP stack. A upload-wide option might be better, but even that might not be the appropriate solution for whatever you're trying to prevent. Let's discuss that in another issue, please.

gap777:
Thanks — that's fair, and the split is done. Closing this in favour of:

- #904 — the service worker fix, and nothing else. `urlStorage.ts` referencing `window` at module scope, which throws a `ReferenceError` in a worker and takes down the browser entry point on import.
- #905 — the `FetchHttpStack` tests, no behaviour change.
- #906 — an issue for the timeout, with the design questions left open rather than a solution presented as settled. No PR attached; happy to implement whichever shape you land on.

PR #904, opened 2026-09-14
Fix urlStorage throwing on import outside a document

gap777:
### Problem

`lib/browser/index.ts` cannot be imported in a Web Worker or a Service Worker. Not "does not report progress" — the import itself fails:

```js
import { Upload } from 'tus-js-client'
// ReferenceError: window is not defined
```

`lib/browser/urlStorage.ts` runs its feature detection at module scope:

```js
hasStorage = 'localStorage' in window
```

A worker has no `window`, so the bare reference throws a `ReferenceError`. The surrounding `catch` only converts `DOMException` (the `SecurityError` from a sandboxed iframe, the `QuotaExceededError` from iOS private mode) and rethrows everything else, so the `ReferenceError` escapes and takes down the whole browser entry point at import time.

The existing comment has the right premise and the inverted conclusion:

```js
// Note: localStorage does not exist in the Web Worker's context, so we must use window here.
```

`window` does not exist there either, which is exactly what makes the line throw. Affects both 4.x and 5.0.0-pre2.

### Fix

`typeof localStorage !== 'undefined'` instead of `'localStorage' in window`. It is `false` where there is no storage, and it still surfaces the `SecurityError` that a sandboxed iframe raises from the property getter, so the private-mode and iframe handling below is unchanged. The storage probe becomes conditional, since it cannot run where there is no storage.

It also checks that `getItem` is actually a function, for runtimes that define the global without a working implementation behind it — Node 25 does this when started without `--localstorage-file`, and the probe would otherwise throw a `TypeError` that escapes the same `catch`.

### Test

The regression test lives in the Node suite, the one environment in this project where `window` is genuinely absent. It imports `lib.esm/browser/urlStorage.js` directly and asserts the module loads, tying `canStoreURLs` to the storage the runtime actually has rather than to a fixed answer — the suite also runs under Deno, which removed `window` but does implement Web Storage.

Verified it still catches the original bug: against the unpatched module, both Node and Deno fail with `ReferenceError: window is not defined`.

Importing the whole browser entry point would be the stronger test, but it hits an unrelated wall first: `is-stream@2` is CJS and Node's native ESM cannot take a named `readable` import from it. Bundlers handle that interop, so it only affects a raw Node import. Happy to widen the test if you would rather solve that too.

### Testing

- Node (jasmine): 80 specs, green
- Browser (karma/puppeteer): 76 specs, green

Split out of #899 as requested. The `FetchHttpStack` tests are in a separate PR, and the `timeout` option will get its own issue first.


gap777:
@Acconut Is this PR more acceptable?

Acconut (approved):
Yes, that looks good, thank you!

gap777:
@Acconut Super! I'd love to rebase my work onto this package once you release it with these changes. 
When can I expect that to happen?

Acconut:
We already merged breaking changes as part of our v5 release in main, so we won't cut another v4 release. I can do another v5 pre-relese soon, but the proper v5 release will take some more time, I'm afraid.

gap777:
@Acconut I was hoping for a v5 pre-release with this SW-compatibility change.

gap777:
Patch release?

Acconut:
Yes, as mentioned a v5 pre-release is possible. I'll try to do it in the next days.

